PR #3, opened 2020-01-15
Fix many broken examples

sjasonsmith:
I attempted to build every example (in an automated fashion) and resolved most of the issues encountered. Every example in this commit was previously unbuildable without further customization of either the pins or config files.

The last commit (the one with miscellaneous fixes) has what I consider the most questionable changes. I had to decide between enabling a new feature (such as advanced park) versus disabling a different feature.

There are still about a dozen examples that do not build, but I couldn't readily determine an appropriate fix for them. Now that this repo exists perhaps I should open an issue for each of them and maybe someone else could help answer questions about them.


thisiskeithb:
I was just thinking of doing something similar. Great work! Creating issues for each broken config would work for me. I can work through them as I have time since I have a feeling I'll be here a lot. 🙂

sjasonsmith:
I am going to delay creating those issues for right now, in case @thinkyhead has any specific guidance for how he wants to handle that in this repo.

Once I get the go-ahead, I can create pretty specific issues with what I found for each one, that will hopefully be easy to address for someone who has actually used the printer.

sjasonsmith:
I apparently didn't mark this as editable by maintainers either. That is fixed now.

Vertabreak:
GitHub suddenly set not editable on me today also was weird i didn't even know there was a option till it was suddenly set without my knowledge.

PR #60, opened 2020-04-04
Malyan M300 / Monoprice Delta Mini config

mojocorp:


thinkyhead:
Brought configs up to date and merged from the command-line.

mojocorp:
@thinkyhead Unfortunately the M300 config does not work and is dangerous. During your rework of the config you forgot to comment out 'ORIG_E0_AUTO_FAN_PIN' meaning the nozzle fan does not work. This fan is also responsible for cooling the extruder and if it's not running, there is a risk of melting it. I'll propose a new PR to fix that.

thinkyhead:
It's part of my job to bring configs up to date when bringing them in, and errors will happen. I would rather have the patch than just the information, because I have a thousand things to do every day.

PR #66, opened 2020-04-11
FlyingBear P905H Single-E configuration

rumatoest:
Tested on my P905H Flying Bear printer with Single Extruder, Dual Z axis and inductive bed level sensor.


thinkyhead:
Thanks for the contribution!

PR #70, opened 2020-04-13
Fix the Malyan M300 configuration

mojocorp:


thinkyhead:
Thanks for the corrections. Much appreciated!

PR #76, opened 2020-04-19
Add support for Daycom 3DP-100

Tho85:
This adds support for the Daycom 3DP-100 printer, sold by German reseller Pollin Electronic. It consists of a Makerlab.me Mega controller, a Mini panel and an MK8 extruder. It does not have a heatbed.

The device orginally ships with a [customized version of Marlin 1.0.2](https://github.com/Tho85/Marlin-Daycom-3DP-100/tree/daycom-3dp-100-original). I have since included all changes made by the manufacturer in this configuration and used it with Marlin 1.1.9 for years. I recently upgraded it to 2.0.5.3, and it works flawlessly (when compiled with [this fix](https://github.com/MarlinFirmware/Marlin/pull/17601)).

There are some improvements over the original version:

* Correctly use the fan as an extruder fan instead of a part fan
* Step and PID parameters have been moved from EEPROM to the configuration
* Better rotary encoder handling
* Enabled S-curve acceleration

### DIFF
--- a/config/examples/Daycom/3DP-100/Configuration.h
+++ b/config/examples/Daycom/3DP-100/Configuration.h
@@ -366,7 +366,7 @@
  *   202 : 200k thermistor - Copymaster 3D
  *     3 : Mendel-parts thermistor (4.7k pullup)
  *     4 : 10k thermistor !! do not use it for a hotend. It gives bad resolution at high temp. !!
- *     5 : 100K thermistor - ATC Semitec 104GT-2/104NT-4-R025H42G (Used in ParCan & J-Head) (4.7k pullup)
+ *     5 : 100K thermistor - ATC Semitec 104GT-2/104NT-4-R025H42G (Used in ParCan, J-Head, and E3D) (4.7k pullup)
  *   501 : 100K Zonestar (Tronxy X3A) Thermistor
  *   512 : 100k RPW-Ultra hotend thermistor (4.7k pullup)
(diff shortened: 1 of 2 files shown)


thinkyhead:
Looks good, thanks!

config/examples/Daycom/3DP-100/Configuration.h
```diff
@@ -737,7 +737,7 @@
  * Override with M92
  *                                      X, Y, Z, E0 [, E1[, E2...]]
  */
-#define DEFAULT_AXIS_STEPS_PER_UNIT   { 80, 80, 800, 94.3 }
+#define DEFAULT_AXIS_STEPS_PER_UNIT   { 80, 80, 4000, 500 }
```
Tho85, line 740: Thanks for merging the PR. :+1:

@thinkyhead Why was this line actually removed? My understanding ist that especially the steps per unit are model-specific and should be included in an example configuration. `{ 80, 80, 800, 94.3 }` are the values provided by the manufacturer and known to work.

thinkyhead: Oops! The configs were not up to date, so I had to update them, and as I do this several times a day in a rush mistakes are bound to happen. To prevent this in future make sure the configs are as up to date as possible so that no fixups will be required by the overworked maintainers.

Tho85: Ah sorry, I just now saw the default configuration in this repository on which I should have based my changes... Thanks for the clarification!

PR #86, opened 2020-04-30
Added some printer based on mightyboards

chepo92:
Some boards have mega1280 and some 2560 so had to cut off some functions to make space for mega1280. 
Flashforge: Creator 2X (wood frame, dual extruder, heated bed), print bed 230x150. - MEGA2560 
QidiTech: Qidi 1, (metal, dual/single extruder, HB), print bed 230x150. - MEGA2560 

Mbot: Cube (wood, dual extr, I think the stock is no Heatbed, but very easy to add and wire, this config uses HB), Print Bed 250x250. - MEGA1280
Wanhao Duplicator 4S , (metal, dual extruder, HB), print bed 230x150. - MEGA1280 

Notice: All examples consider the origin (0,0) in the origin (aka left front), as the original config with origin at center is kind of deprecated for cartesians printers.


thinkyhead:
Some of these are way out of date (Mbot). Because of this it is hard to tell when a setting (e.g., `PID_FUNCTIONAL_RANGE`) is:
- intentionally different from the default setting, or
- there is just a new default, and the config has the old default.

I will do my best to patch them up, but I am probably going to make mistakes so please review.

PR #87, opened 2020-04-30
Geeetech Rostock 301 patches

chepo92:
will add new thermistor table
extruders where inverted
Z safe homing should be at center 0,0
slow homing please!
relaxed thermal runaway time


config/examples/delta/Geeetech/Rostock 301/Configuration.h
```diff
@@ -1468,12 +1470,12 @@
...
 #if ENABLED(Z_SAFE_HOMING)
-  #define Z_SAFE_HOMING_X_POINT ((X_BED_SIZE) / 2)    // X point for Z homing when homing all axes (G28).
-  #define Z_SAFE_HOMING_Y_POINT ((Y_BED_SIZE) / 2)    // Y point for Z homing when homing all axes (G28).
+  #define Z_SAFE_HOMING_X_POINT X_CENTER  // This is a delta and should be at center (0,0) (G28).
+  #define Z_SAFE_HOMING_Y_POINT Y_CENTER  // This is a delta and should be at center (0,0) (G28).
 #endif
 
 // Delta only homes to Z
-#define HOMING_FEEDRATE_Z  (200*60)
+#define HOMING_FEEDRATE_Z  (1000)   // 200*60 was killing the endstop
```
thinkyhead, line 1478: The maximum attainable speed for Z should typically be used for the homing speed. If the max feedrate for Z is set too high then this fix alone won't prevent other fast and potentially damaging Z moves. So you may want to augment this by also limiting the Z max feedrate.

PR #88, opened 2020-05-01
Fix Tornado bed width

fxwgr:
Point 3, 6 and 9 are outside of the bed because of wrong bed size set.
Tevo Tornado's printing range is 300x300x400.

### DIFF
--- a/config/examples/Tevo/Tornado/V1 (MKS Base)/Configuration.h
+++ b/config/examples/Tevo/Tornado/V1 (MKS Base)/Configuration.h
@@ -1107,7 +1107,7 @@
 // @section machine
 
 // The size of the print bed
-#define X_BED_SIZE 320
+#define X_BED_SIZE 300
 #define Y_BED_SIZE 300
 
 // Travel limits (mm) after homing, corresponding to endstop positions.
--- a/config/examples/Tevo/Tornado/V1 (MKS Base)/Configuration.h
+++ b/config/examples/Tevo/Tornado/V1 (MKS Base)/Configuration.h
@@ -1107,7 +1107,7 @@
 // @section machine
 
 // The size of the print bed
-#define X_BED_SIZE 300
+#define X_BED_SIZE 310
 #define Y_BED_SIZE 300
 
 // Travel limits (mm) after homing, corresponding to endstop positions.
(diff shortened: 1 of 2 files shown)


thisiskeithb:
Advertised is 300x300x400, but you could print 310mm on X unless that changed on later models.

fxwgr:
Yes, but I tried it with these 320mm and even the bed leveling is not working correct on point 3, 6 and 9. The nozzle is outside of the bed. That clearly shouldn't do so.
I've got the orange model of the Tornado with the glass bed.
This are the measures of my default bed: 310mm (x) x 370mm (y)
Therefore 320mm bed size on x is too much in my case.

thisiskeithb:
I probably didn't verify on my Tornado when I brought them over from Tevo.

Edit: The V2 should be updated as well:
https://github.com/MarlinFirmware/Configurations/blob/1abfdab06e0a922187cafc06f1d5e20adf1c8a46/config/examples/Tevo/Tornado/V2%20(MKS%20GEN-L)/Configuration.h#L1109-L1111

thinkyhead:
Please submit any patches to the `import-2.0.x` branch.

fxwgr:
@thisiskeithb I don't exactly know if the V2 maybe has another bed and therefore another bedsize?

@thinkyhead am I wrong or am I in the right branch?
The line below the title of the pull request says:
> merge 1 commit into MarlinFirmware:import-2.0.x

thinkyhead:
Looks good. For some reason I thought this page was a plain issue and not a PR.

thinkyhead:
So anyway… If the nozzle can reach an area that is 320mm wide then the bed size should be set up to accommodate that. The bed size should not be constrained simply because of a probe offset. Instead, either use UBL which permits filling in points where the probe can't reach, or use ABL which probes only where it can reach and extrapolates beyond that.

fxwgr:
I don't exactly get, why the defined bed size should exceed the real bed size. Are there any advantages?

You mean instead the UBL/ABL shouldn't refer anymore to the defined bed size and set to the actual real bed size? Did I understand that right?

thinkyhead:
> I don't exactly get, why the defined bed size should exceed the real bed size.

It should not. But Keith implied that the stock bed has 320mm of usable width.

thisiskeithb:
It should be 310mm on X. I just measured the stock bed on my Tornado.

fxwgr:
Okay, I adjusted it to 310mm to get the maximum bed size and Tornado v2 also.

PR #101, opened 2020-05-11
Increase Z Hybrid Threshold in SKR Mini E3 configs

thisiskeithb:
## Description

Increased `Z_HYBRID_THRESHOLD` values since SpreadCycle was kicking in too early.

### DIFF
--- a/config/examples/BigTreeTech/SKR Mini E3 1.0/Configuration_adv.h
+++ b/config/examples/BigTreeTech/SKR Mini E3 1.0/Configuration_adv.h
@@ -2348,7 +2348,7 @@
   #define X2_HYBRID_THRESHOLD    100
   #define Y_HYBRID_THRESHOLD     100
   #define Y2_HYBRID_THRESHOLD    100
-  #define Z_HYBRID_THRESHOLD       3
+  #define Z_HYBRID_THRESHOLD      30
   #define Z2_HYBRID_THRESHOLD      3
   #define Z3_HYBRID_THRESHOLD      3
   #define Z4_HYBRID_THRESHOLD      3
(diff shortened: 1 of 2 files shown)


swilkens:
@thisiskeithb I would suggest to disable `#HYBRID_THRESHOLD` by default on this board.

I see a fairly large lump of issue reports and threads on Reddit with people wondering why their "silent" board is making noises after a firmware upgrade. 9/10 People are buying this board for silent operation.

I think it's good to disabled `#HYBRID_THRESHOLD` by default, the few that actually want it can then configure it manually.

thisiskeithb:
What about the potential for lost steps without the ability to switch to SpreadCycle?

swilkens:
> What about the potential for lost steps without the ability to switch to SpreadCycle?

I'm not losing steps at the stock printing speed / acceleration on this printer and 50 mm/s extruder, with this board in the stock configuration. It may become an issue when users start to add weight to the X, Y carriage (i.e. direct drive, glass bed)... but then it's already customized, they can look at features to compensate?

Honestly; just trying to reduce the incoming "why is my printer making noise with this silent board" bug report flood - given that people typically buy this to enable silent operation.

Alternatively, bump up E to 50 as well? May already mitigate things enough without risking missed steps on setups with glass bed / direct drive, which I suppose are fairly popular.

thisiskeithb:
I'll submit a new PR.

thisiskeithb:
@swilkens: See https://github.com/MarlinFirmware/Configurations/pull/103 for disabling `HYBRID_THRESHOLD `.

swilkens:
If we now get a flood of missing steps reports, feel free to slap me.

thinkyhead:
> I'll submit a new PR.

You can always add more commits to an existing PR and change the title, if you prefer.

PR #103, opened 2020-05-11
Disable Hybrid Threshold on SKR Mini E3 configs

thisiskeithb:
## Description

Disable `HYBRID_THRESHOLD` on SKR Mini E3 configs because users aren't used to the noise when TMCs kick over to SpreadCycle.

This should be fine for a stock Ender-3, but users can always enable & tune to their liking if they print at higher speeds, go direct drive, install a glass bed, etc.

See comments in https://github.com/MarlinFirmware/Configurations/pull/101 for more info.

### DIFF
--- a/config/examples/BigTreeTech/SKR Mini E3 1.0/Configuration_adv.h
+++ b/config/examples/BigTreeTech/SKR Mini E3 1.0/Configuration_adv.h
@@ -2342,7 +2342,7 @@
    * STEALTHCHOP_(XY|Z|E) must be enabled to use HYBRID_THRESHOLD.
    * M913 X/Y/Z/E to live tune the setting
    */
-  #define HYBRID_THRESHOLD
+  //#define HYBRID_THRESHOLD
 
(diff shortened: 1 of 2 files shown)


sjasonsmith:
I know I have already done this! It must have gotten lost in the shuffle when examples were all moved to the other repo.

qwewer0:
<s>Won't it cause skipped steps and stepper driver overheating at high (XY > 100mm/s) speeds on StealthChop mode?</s>

Edit: Well, if we get a lot of missed steps reports, then we can always slap @swilkens.

thisiskeithb:
> Won't it cause skipped steps and stepper driver overheating at high (XY > 100mm/s) speeds on StealthChop mode?

You may need to enable `HYBRID_THRESHOLD` and tune it if you're printing that fast. Cura & PrusaSlicer profiles for the Ender-3 are 1/2 that or less. Also:

> See comments in #101 for more info.

PR #105, opened 2020-05-15
Added Ender-3 Pro

Penguin096:
Logo (Statusscreen.h), Machine Name (Configuration.h)


qwewer0:
The change in `DEFAULT_AXIS_STEPS_PER_UNIT` from `80, 80, 400, 93` to `79.75, 79.75, 400, 98.4` is intentional? Shouldn't the Pro have the same XY steps/mm as the non pro one?

thinkyhead:
Yes. The default steps-per-unit should be set according to the mechanical parts, not according to user tuning. An increase in E steps is typical, because there tends to be slippage and users try to print too fast. The extruder steps should be calibrated while cold and with no resistance on the filament.

qwewer0:
Ok, so apart the E steps/mm the X and Y steps/mm is different on the Ender 3 and Ender 3 Pro?

qwewer0:
@thinkyhead If the 79.75 steps/mm for X and Y axis is more precise and closer to the real hardware values, then shouldn't the Ender 3 be the same as the Pro?

thinkyhead:
> If the 79.75 steps/mm for X and Y axis is more precise

I will defer to a panel of 10 or more experienced 3D printer users on what the correct settings should be, and will happily accept their corrections in the form of a PR.

AnHardt:
While 80 steps/mm is the theoretical value for tension defined by the maker of the belt, 79.75 is explainable for a belt stretched a bit more.  As long there is no rule for tensioning the belt defined by the makers of the printer and everyone can stretch it to his taste, a deviation of a bit more or less then 1/4% is meaningless as a default value. At the end everyone who wants to make a finetuning for a special machine has to find out the best value by himself. (This is different for spindle driven  systems.)

thinkyhead:
I tend to favor the "idealized" setting based on the belt / pulley pitch. Even if a belt is stretched, only the extreme ends of the X moves would be affected (i.e., backlash / slop). The steps within the line should all be in lock-step with the motor at 80 steps/mm.

AnHardt:
I agree with setting the nominal belt/pulley values as defaults.
I disagree with your model of what happens in a belt-system under extra tension.

Let my try to explain without a drawing.
Let's assume a belt with 2 mm pitch at nominal tension and matching pulleys with 20 teeth.
The centers of the pulleys shall be 200 mm away from each other. So we need a belt with 220 teeth, On each side 10 teeth are meshed with the pulleys and the free upper and lower belt between the pulleys is 100 teeth long.
Now we stretch the belt to get a bit more tension. For that we increase the distance between the centers of the pulley by 1% = 2 mm. The free length is now 2 * 202 mm but still 100 teeth.
That's idealized. At the pulleys it's a bit more complicated. 
On the one side, the belt has to stretch also in the area with contact to the pulley. On the other, the teeth have to mesh with the grooves in the pulley. Additionally symmetry asks for the the upper half of the system must be like the lower.  Now we can assume the tee at the 9 o-clock position on the left and the 3 o-clock position at the right see nominal stress and fill the gaps in the pulleys perfectly. One tee up and down sees a bit more stress and deforms a bit. That deformation allows a tiny bit of displacement of the belts back and the next tee. The next tees up and down see a bit more stress, are deformed a bit more, ... , until we are at the last meshed tees what are deformed most to fit into the grooves, seeing the full stress of the stretched belt. That model works well until either the belt tears or the last meshed tee can't deform enough and shears off or the tee of the belt begins to ride on the tee of the pulley, not slipping into the groove. Normally all this is not a problem because the deformations are small and the profile of the teeth an groves is made for that.

No forces are brought into the belt where it is not meshed. Where it is not meshed the force in the belt is constant. Constant force means constant pitch of the teeth. When the force is not nominal the pitch is not nominal. Even in a stretched belt the pitch is even, just not where it is meshed with the pulleys, there it is nominal.

A full rotation of the pulley will still transport 20 teeth but this 20 teeth will not be 40 mm anymore but 40.4 mm.  You have to set a smaller steps/mm.

PR #107, opened 2020-05-18
Disable ARC_SUPPORT

thisiskeithb:
### Description

None of the major slicers support `ARC_SUPPORT`, so let's disable it and save ~3226 bytes/build.

### Benefits

Smaller builds.

### DIFF
--- a/config/default/Configuration_adv.h
+++ b/config/default/Configuration_adv.h
@@ -1649,7 +1649,7 @@
 //
 // G2/G3 Arc Support
 //
-#define ARC_SUPPORT                 // Disable this feature to save ~3226 bytes
+//#define ARC_SUPPORT                 // Disable this feature to save ~3226 bytes
 #if ENABLED(ARC_SUPPORT)
   #define MM_PER_ARC_SEGMENT      1 // (mm) Length (or minimum length) of each arc segment
   //#define ARC_SEGMENTS_PER_R    1 // Max segment length, MM_PER = Min
--- a/config/examples/3DFabXYZ/Migbot/Configuration_adv.h
+++ b/config/examples/3DFabXYZ/Migbot/Configuration_adv.h
@@ -1648,7 +1648,7 @@
 //
 // G2/G3 Arc Support
 //
-#define ARC_SUPPORT                 // Disable this feature to save ~3226 bytes
+//#define ARC_SUPPORT                 // Disable this feature to save ~3226 bytes
 #if ENABLED(ARC_SUPPORT)
   #define MM_PER_ARC_SEGMENT      1 // (mm) Length (or minimum length) of each arc segment
   //#define ARC_SEGMENTS_PER_R    1 // Max segment length, MM_PER = Min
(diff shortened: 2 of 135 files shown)


sjasonsmith:
Seems reasonable to me, especially for anything with only 128KB of flash.

foosel:
Just FYI, @FormerLurker is currently working on a tool that compiles existing GCODE files to *do* make use of arcs, for better throughput via serial, and it would hamper his awesome work *severely* if this was now merged and applied generally.

See this forum thread on his work and promising results, also from the testers from the community: https://community.octoprint.org/t/new-plugin-anti-stutter-need-testers/18077

thisiskeithb:
This wasn’t removing the feature, just disabling it by default.

foosel:
> This wasn’t removing the feature, just disabling it by default

Disabling it by default means practically removing it since the majority of 3d printer owners out there will never compile their own firmware (if they even know what that is in the first place) and just take whatever their printer vendor supplies them with (which usually is some lazily cobbled together config with default settings and just some printer specific parameter changes, possibly with some disabling of safety features sprinkled in to work around cheap ass components). And those who do compile their own will more often than not also never question the defaults.

Thank you for understanding :)

InsanityAutomation:
FWIW I do end up needing to disable this to fit bilinear on Melzi boards nowadays. Its quite a fight to fit a workable build on those. Trade-offs are inevitable at that point unfortunately.

thinkyhead:
For consistency we'll disable it in the config examples that apply to small boards, or models of printer that don't come with arcs enabled as a standard. For any board / machine where the code can fit it should be kept enabled.

thinkyhead:
NOTE: Marlin's handling of many short segments is better these days thanks to improvements in the JD algorithm, but stutters are still certainly possible due to bandwidth. A plugin that can convert curves into arcs will be massively welcome and should make a great improvement in performance and print results.

@FormerLurker — For Marlin 2.0.6 and up you will be able to have the plugin send `M115` to Marlin to see if it has `ARC_SUPPORT` enabled. For earlier versions you can send a "null" `G2`/`G3` command and see if the machine replies "unknown command." If the firmware doesn't support arcs, throw up a bright red and yellow box explaining how to enable it in Marlin.

In-general, the G2/G3 arc commands have been somewhat broken in Marlin and in recent versions the segment lengths are sometimes calculated wrong. So, really, it is only with 2.0.6 and up that you can begin to trust arcs to be a reliable improvement over linear segments.

My recommendation would be to make 2.0.6 the minimum supported version and put up a warning for older versions of Marlin. If you find that certain versions are good at arcs, those could be added to your "trusted" list….

FormerLurker:
@thinkyhead, that would s fantastic, thank you for letting me know!  I will add arc support testing ASAP!

FormerLurker:
> A plugin that can convert curves into arcs will be massively welcome and should make a great improvement in performance and print results

Do you mind if I include this qoute in the plugin description?  I would really appreciate it.

thinkyhead:
> Do you mind…?

Not at all, that's fine.

PR #111, opened 2020-05-24
Initial BTT SKR Mini E3 V2 config

thisiskeithb:
### Description

Initial [SKR Mini E3 V2](https://www.biqu.equipment/collections/skr-series/products/bigtreetech-skr-mini-e3-v2-0-32-bit-control-board-integrated-tmc2209-uart-for-ender-3) config.

### Benefits

SKR Mini E3 V2 support.

### Related Issues

~~https://github.com/MarlinFirmware/Marlin/pull/18088 needs to be tested/merged first.~~ Merged!

PR #115, opened 2020-05-28
Add config(and config adv) examples for MEEB_3DP board

ccccmagicboy:
ender3 and cr10mini example include


thinkyhead:
We don't curate configurations for boards or for modification kits, but only for shipping machine models. Of course, you should feel free to distribute these custom configurations with your board and any kits that include your board.

ccccmagicboy:
👌

ccccmagicboy:
Can this branch continue?

thinkyhead:
I am on the edge of considering this (and all others like it) a problem of clutter. We might have to make a separate place for configs that are related to boards.

The thing is, your board is supposed to work with any 3D printer as long as the right components are plugged into the right pins, so giving examples makes it look like this repo is being used for marketing.

thisiskeithb:
> The thing is, your board is supposed to work with any 3D printer as long as the right components are plugged into the right pins, so giving examples makes it look like this repo is being used for marketing.

Have you received complaints about this?

I’m having a difficult time seeing the difference between hosting board configs (usually with specific size/electrical constraints targeted at specific machines) vs. branded configs for various printer manufacturers.

I look at it as a great way to keep configs (of both types) up to date so users don’t have to rely on manufacturers to keep their forks/sources up to date, since many manufacturers don’t.

ccccmagicboy:
Hi, I am the designer of the meeb 3dp board, and the printer configuration file examples I provide here are mainly used to test my meeb 3dp controller board for a better experience for awesome marlin firmware. Because there are so many settings for the marlin firmware, and the update is very fast at the same time, I will only focus on the test on my existing printer(ender3, cr10mini, etc), the configuration file will continue to be updated in my repos, and if the files could be submitted here, I Also willing to submit here.

thinkyhead:
We _could_ make a "variants" sub-folder for _every printer_ and add configurations for every possible board+printer combination, but we don't want to go down that road. A few boards were tossed into the config examples mix during 32-bit development and this arrangement is not beloved.

I recognize that now a lot of boards are appearing which capitalize on the Creality Ender series and present a matching form-factor. I want to make some kind of room for these when they require special settings (for example, when some pins are inverted compared to normal, or if the board has different stepper drivers). However, I don't want to make the board name the top-level folder for these examples.

So I propose that we put all configs like this within the config folders for the printers that they belong to.

I will do some re-arrangement and push the changes soon….

thisiskeithb:
@thinkyhead: Where should generic board configs like those for MKS go? https://github.com/MarlinFirmware/Configurations/tree/import-2.0.x/config/examples/Mks

There are still board configs (some for specific printers, some not) from Fysetc: https://github.com/MarlinFirmware/Configurations/tree/import-2.0.x/config/examples/FYSETC

I can dig through the rest of the configs, but those are some of the more popular ones I’m aware of.

thisiskeithb:
@thinkyhead: I noticed the SKR Mini E3 V2.0 config from PR https://github.com/MarlinFirmware/Configurations/pull/111 is missing now as well: https://github.com/MarlinFirmware/Configurations/tree/import-2.0.x/config/examples/Creality/Ender-3

Any plans to bring that back?

thinkyhead:
Sorry, usually I manage to not overwrite `import-2.0.x` by mistake. Let me see if I can revive that history.

PR #118, opened 2020-05-31
JGAurora A3

Dingo420:
Copied code from 1.1 to work with 2.0


VinnyCordeiro:
For clarification, this PR is for a JGAurora A3 3d printer, which does not have an example configuration file available. I provided the means for OP to successfully compile a newer version of Marlin for his old printer.

PR #119, opened 2020-06-01
SKR mini E3 1.2: Stay close to the default config

TheZoker:
### Description

For this PR I tried to reduce the changes for the SKR mini E3 1.2 board to the minimum required to run with the board. Before there where many custom values included, that (in my opinion) should not be part of the default config. 
The config is based on the ender-3 config with necessary changes for the SKR mini E3 1.3

@thisiskeithb Can you please have a look at this PR and tell me if your fine with this? 

### Benefits

Have a more clean and minimal default config for this board

### Related issues

none


thisiskeithb:
I think they were fine how they were. I don't understand many of the changes here.

thisiskeithb:
The updated board allows more features to be enabled and you've enabled features that won't compile unless you copy out files from the Ender-3 config folder. Also, some of these are heavily used or requested by users.

If you prefer the stock config, I'd recommend starting with that instead of modifying this one (you'd have to modify all the E3 Mini configs to match for consistency anyway).

config/examples/BigTreeTech/SKR Mini E3 1.2/Configuration_adv.h
```diff
@@ -1567,9 +1565,9 @@
 #define LIN_ADVANCE
 #if ENABLED(LIN_ADVANCE)
   //#define EXTRA_LIN_ADVANCE_K // Enable for second linear advance constants
-  #define LIN_ADVANCE_K 0.0    // Unit: mm compression per 1mm/s extruder speed
+  #define LIN_ADVANCE_K 0.22    // Unit: mm compression per 1mm/s extruder speed
```
thisiskeithb: This should not be enabled by default. Users need to tune this setting on their own, which is why it's left at 0.0.

TheZoker: The reason for reverting it to 0.22 was, that this is the default value:
https://github.com/MarlinFirmware/Configurations/blob/import-2.0.x/config/default/Configuration_adv.h#L1569

Should this value than also changed in the default file?

TheZoker:
While it's true, that the board allows for more features (such as linear advance), some things like `EXPERIMENTAL_SCURVE` should not be enabled by default (in my opinion). There is a check, that prohibits linear advance and s curve to be used together and for a default config, I think this check should not be bypassed by enabling an experimental feature. 

>  I don't understand many of the changes here.

What changes to you not not understand? Basically I just reverted a lot of custom values that need to be changed anyway or are just experimental stuff. 

And yes my plan was to change also version 1.0 of the board, when the changes are discussed in this PR. 

> Also, some of these are heavily used or requested by users.

Which features are you referring to?

If these features are interesting for many people, I think the right would be to change the base config or the ender 3 config and not include them in this specific board.

thisiskeithb:
Like I said, I think they were fine how they were.

I build and maintain these & the rest of the configs for BigTreeTech and am very active in the BigTreeTech communities.

I don’t think they should be changed.

PR #120, opened 2020-06-02
Anycubic Mega Zero Marlin 1.1.9 configs, based on Anycubic's v0.0.4 FW

kad:
Config files are re-created based on the source code published at https://github.com/ANYCUBIC-3D/MEGA_ZERO


thinkyhead:
Thanks! Since 1.1.x gets no more releases, this is the right target for 1.1.9 PRs.

PR #121, opened 2020-06-02
Anycubic Mega Zero Marlin 2.0.5 configs, based on Anycubic's v0.0.4 FW

kad:
Config files are re-created based on the source code published at https://github.com/ANYCUBIC-3D/MEGA_ZERO


thinkyhead:
Thanks for your contribution! Unfortunately we can't accept PRs directed at release branches. We make patches to the "`import`" branches and only later do we push them out as releases. Please redo this PR starting with the `import-2.0.x` branch and be careful to target `import-2.0.x` when resubmitting the PR.

thinkyhead:
Actually, I don't mind pitching into the 2.0.5 pile with this. But I'll also have to add this into `import-2.0.x` because there's no automatic interchange from here to there.

kad:
thanks @thinkyhead !
BTW, it might be good to create Contributing.md for this repository, to show where and which patches are expected, and policies for patches to be accepted.

thinkyhead:
Thanks for the suggestion. I will add one soon.

PR #122, opened 2020-06-03
Renkforce RF100: Update heater and bed max. temperature envelope according to data sheets

dok-net:
Commit ef83d0e88b1c518861510a2e17618fc424b8fd60 switched to using the #define BED_MAXTEMP  to set  PTC_MAX_BED_TEMP, but failed to update the former as well.
This is fixed by this PR. Additionally, the maximum temperature for the heater is lower, according to product data sheets. In fact, operation above 230°C is only allowed for up to 10min at a time, but 260°C is listed as possible, so for specialty uses, like PLA wood resin, the higher value is allowed.

### DIFF
--- a/config/examples/Renkforce/RF100/Configuration.h
+++ b/config/examples/Renkforce/RF100/Configuration.h
@@ -462,15 +462,15 @@
 // Above this temperature the heater will be switched off.
 // This can protect components from overheating, but NOT from shorts and failures.
 // (Use MINTEMP for thermistor short/failure protection.)
-#define HEATER_0_MAXTEMP 275
-#define HEATER_1_MAXTEMP 275
-#define HEATER_2_MAXTEMP 275
-#define HEATER_3_MAXTEMP 275
-#define HEATER_4_MAXTEMP 275
-#define HEATER_5_MAXTEMP 275
-#define HEATER_6_MAXTEMP 275
-#define HEATER_7_MAXTEMP 275
-#define BED_MAXTEMP      150
+#define HEATER_0_MAXTEMP 260
+#define HEATER_1_MAXTEMP 260
+#define HEATER_2_MAXTEMP 260
+#define HEATER_3_MAXTEMP 260
+#define HEATER_4_MAXTEMP 260
+#define HEATER_5_MAXTEMP 260
+#define HEATER_6_MAXTEMP 260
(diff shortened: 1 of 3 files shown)


thisiskeithb:
Those may have been set to 275 to get around Marlin’s 15 degree safety margin for hotends. By setting max temp to 260, you’ll only be able to set the temp to a max of 245.

dok-net:
@thisiskeithb Can you please point me at the documentation of this 15K safety margin? I'm unsure if, for a ready-to-run device, anything faintly dangerous should be in a default firmware configuration, so I guess I'd rather figure out how to prevent any temperature that causes damage to the device, even if it takes 10min to do so.

dok-net:
@thisiskeithb OK, some Googling later, found only another person's opinion on the 15K margin, but I'm leaving the 260°C max as a compromise between safety (damage after 10min of sustained high temperature) and utility.
By the way, the original 275 are unlikely by special consideration for the RF100s, but rather the template file's default. I for one, being the submitter of the current RF100 configurations, never noticed or touched these particular settings.
Researching the original firmware as shipping documentation, they allow setting up to 260°C, albeit under the aforementioned precautions. Considering that the PTFE tube in the hotend isn't stable above 250°C according to PTFE material specs, the max 245°C seems a safer choice, no matter what.

thisiskeithb:
> @thisiskeithb Can you please point me at the documentation of this 15K safety margin?

Search the code for `HOTEND_OVERSHOOT`Here's where it's [defined](https://github.com/MarlinFirmware/Marlin/blob/2d1cbf8e1d9f4d55f79b5c5d0628965472a7f92e/Marlin/src/inc/Conditionals_LCD.h#L471-L475) and then applied in [temperature.h](https://github.com/MarlinFirmware/Marlin/blob/2d1cbf8e1d9f4d55f79b5c5d0628965472a7f92e/Marlin/src/module/temperature.h#L591-L593) (it's used in several other places as well).

Similarly, the bed's max temp is [`BED_MAXTEMP`-10](https://github.com/MarlinFirmware/Marlin/blob/2d1cbf8e1d9f4d55f79b5c5d0628965472a7f92e/Marlin/src/inc/Conditionals_post.h#L1723-L1726).

> max 245°C seems a safer choice, no matter what.

That’s fair 🙂

PR #123, opened 2020-06-03
Use PROBING_MARGIN instead of MIN_PROBE_EDGE

thisiskeithb:
### Description

Use `PROBING_MARGIN` instead of `MIN_PROBE_EDGE`.

### Benefits

SKR Mini E3 V2 & JGAurora A3 configs compile without errors.

### Related Issues

https://github.com/MarlinFirmware/Marlin/pull/18140

### DIFF
--- a/config/examples/BigTreeTech/SKR Mini E3 2.0/Configuration.h
+++ b/config/examples/BigTreeTech/SKR Mini E3 2.0/Configuration.h
@@ -969,7 +969,7 @@
 
 // Most probes should stay away from the edges of the bed, but
 // with NOZZLE_AS_PROBE this can be negative for a wider probing area.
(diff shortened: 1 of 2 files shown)


thisiskeithb:
Found the JGAurora A3 config was also broken, so I included that in this PR.

alvinlgs:
Hi there. Seems like configuration.h for SKR Mini E3 1.2 also requires the same fix. Using the latest bugfix2.0.x configuration files. Example below.

_// Most probes should stay away from the edges of the bed, but
// with NOZZLE_AS_PROBE this can be negative for a wider probing area.
#define MIN_PROBE_EDGE 17 //alvinlgs 10_

PR #126, opened 2020-06-06
Move more Ender-3 configs

thisiskeithb:
### Description

Following up to the aftermarket board config move discussed in https://github.com/MarlinFirmware/Configurations/pull/115, this PR moves the ~~BigTreeTech SKR Mini E3 V2~~ (moved in 86e9bbe418778c17d58aa43e188b21586cacf791 after this PR) and FYSETC Cheetah configs to the Creality/Ender-3 folder.

### Benefits

More consistent application of the new aftermarket board rules 🙂 

### Related Issues

https://github.com/MarlinFirmware/Configurations/pull/115


thisiskeithb:
What should happen to the board-only configs like those still located in the [FYSETC](https://github.com/MarlinFirmware/Configurations/tree/import-2.0.x/config/examples/FYSETC), [MKS](https://github.com/MarlinFirmware/Configurations/tree/import-2.0.x/config/examples/Mks), [STM32](https://github.com/MarlinFirmware/Configurations/tree/import-2.0.x/config/examples/STM32), [UltiMachine](https://github.com/MarlinFirmware/Configurations/tree/import-2.0.x/config/examples/UltiMachine), etc. folders?

thinkyhead:
> What should happen to the board-only configs…?

These will have to be examined to decide what to do with each. The default configuration for Marlin is still an old Ultimaker, so that's not a very suitable demo model for boards.

If a board config isn't targeted to a particular type of machine, we can put them under "cartesian/generic" and then just add a README to describe what kind of basic machine they are implementing.

Really, we only need to have examples for boards when something in the configuration has to be changed in some special way that isn't already covered in the board itself, or when there needs to be extra documentation around certain options. Usually this will be something with Trinamic drivers or serial ports.

I'm considering moving all the configs not in DELTA/SCARA into a "Cartesian" top-level folder. And, also considering splitting up the Cartesial models based on their kinematics. Such as Mendel style, Core/H-Bot style, and others.

So the folder hierarchy would be `type/subtype/vendor/model/variant`…

withinboredom:
@thinkyhead 
Why were the configs changed to Japanese?

thisiskeithb:
Keep in mind that character set != lcd language.

The charset was reverted since it doesn’t apply to full graphic displays. The language is still English.

withinboredom:
Thanks!

PR #129, opened 2020-06-13
Zonestar P802M Support

perivar:
Support for Zonestar Prusa P802M which is quite similar to Anet A8 or one of the Troxy versions.
Note! This uses the new zonestar thermistor 502 I added to the Marlin firmware bugfix.


thinkyhead:
Thanks for your contribution! Unfortunately we can't accept PRs directed at generated branches. We make patches to the import branch and use it to generate the other branches. Please redo this PR starting with the `import-2.0.x` branch and be careful to target `import-2.0.x` when resubmitting the PR.

thinkyhead:
I've created a new one at #130 so you're all set!

PR #131, opened 2020-06-17
Added CTC i3 GT2560 Rev A clone configuration without probe

madmachinations:
### Requirements

### Description

Adds a working example configuration for the CTC "Prusa i3" DIY printer running on a clone of the GT2560 Rev A board. PID and step settings remain the same as those set in the factory.

I did look at the current GT2560 Rev A example configuration but the settings it ships with did not match the requirements of the CTC.

### Benefits

The original manufacturer does not ship a copy of the Marlin firmware they installed at the factory, nor do they host a copy to download. So I had to put this together with what I could get the original firmware to tell me combined with trial and error.

Having this example configuration enables others with this clone kit printer to be able to upgrade from the boards original Marlin 1.0 while retaining it's factory calibration.

### Related Issues

This is not related to any existing or past issue that I could see in the issue tracker.


thinkyhead:
I'm reluctant to include this because we don't even have a standard config.
What is the URL of this printer's official website?

madmachinations:
This is their website [https://www.ctcprinter.com/](https://www.ctcprinter.com/)

While they make many clone printers, unfortunately there are no firmware downloads available on there.

The [support](https://www.ctcprinter.com/download.php?CateId=2) section lists one download, but:

- It isn't applicable to this or indeed most of their printers.
- It isn't actually a link, there's no working download. It's just some text.

:upside_down_face:

madmachinations:
Up to you if you want to merge it or not :man_shrugging: . It looks like a bunch of the settings have been overwritten by the update. If you want to go ahead then I can put them back into the version you've brought it up to.

thinkyhead:
> It looks like a bunch of the settings have been overwritten by the update.

Nothing was changed. See b14b437

thinkyhead:
Ah, I see. A "rename" of folders reverted the config file… Now fixed.

madmachinations:
Ah, thanks :)

PR #134, opened 2020-06-21
Creality Ender 5 Pro Config Updates

boxman0617:
### Requirements

### Description

- Fixed error w/ X_HOME_BUMP_MM not being found

When first compiling the Ender 5 Pro configuration files for the latest Marlin 2.0.5.3.
A quick search lead me to this: https://github.com/MarlinFirmware/Marlin/issues/17594
So I fixed it in this configuration file

- Configured home to match factory bottom left

After flashing Marlin 2.0.5.3 to my Ender 5 Pro with the configurations I am attempting to change, factory homing all of the axes would result in the correct position after hitting the endspots, but after prints where completed the ender would travel to the opposite side (top right).

These changes make it so the X and Y Steppers are NOT inverted and changes the direction of endstops when homing to MIN so we would still home to the factory default (bottom left) and also everything is printed in the correct orientation and the home position is correct on competition.

- Updates to Preheat Constants to match standard temperatures

I have noticed a lot of PLA plastics being set to these temperatures for this machine, so I changed them as I had to keep editing it manually.

### Benefits

1. Can compile Marlin 2.0.5.3 successfully for Ender 5 Pro
1. Home and orientation is set to factory defaults
1. Standard PreHeat Constants for Ender 5 Pro are used

### Related Issues

https://github.com/MarlinFirmware/Marlin/issues/17594

### DIFF
--- a/config/examples/Creality/Ender-5 Pro/Configuration.h
+++ b/config/examples/Creality/Ender-5 Pro/Configuration.h
@@ -632,11 +632,11 @@
 // Specify here all the endstop connectors that are connected to any endstop or probe.
 // Almost all printers will be using one per axis. Probes will use one or more of the
 // extra connectors. Leave undefined any used for non-endstop and non-probe purposes.
-//#define USE_XMIN_PLUG
-//#define USE_YMIN_PLUG
+#define USE_XMIN_PLUG
+#define USE_YMIN_PLUG
 #define USE_ZMIN_PLUG
-#define USE_XMAX_PLUG
-#define USE_YMAX_PLUG
+//#define USE_XMAX_PLUG
+//#define USE_YMAX_PLUG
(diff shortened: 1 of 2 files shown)


thinkyhead:
The configuration here is not intended to be compatible with 2.0.5.3.

thinkyhead:
We'll need confirmation and signoff from the original creator of MarlinFirmware/Marlin#16221 which is @thisiskeithb .

thisiskeithb:
These changes aren’t correct since Creality ships the Ender-5 with Ender-3 settings, which sets 0,0 to the back right corner, not the front left corner. The machine homes to both X and Y max (where the endstops are located), not min.

This has been discussed in the main Marlin repo before (I'm failing to find the thread at the moment), but we should be providing configs with proper settings.

PR #147, opened 2020-07-08
Remove no more valid message

GMagician:


thinkyhead:
Wrong target. Should go to the `import-2.0.x` branch. I tried to change the base but all it did was close the sucker.

GMagician:
> Wrong target

That's my first time and I'm gone by how I was used to...

PR #159, opened 2020-07-21
Added Creality Ender 5 with SKR 1.4 (Turbo) configs

jagoosw:
Confi file also has options for

### Requirements

Needed default configs for Creality Ender 5 with SKR 1.4 (Turbo)  and options for all metal hotend and chamber configs.

### Description

I have added a set of files for the Creality Ender 5 with SKR 1.4 Turbo.

I have also made the config file easy to change for the following options:
- TMC2209 drivers, I will add others at a later date, probably more efficiently
- E3D all metal hotend - changes thermistor and max temps
- Extreme mode for all metal hotend - this sets the max temp higher which is not recommended for long term use by E3D
- Chamber setup - configures chamber with the thermistor in the E1 thermistor pin on the SKR 1.4 and the heater in the second heater pins beside the bed heater.

### Benefits

This achieves the requirements making it easier to configure the Ender 5 for an SKR 1.4 (Turbo). When I was setting it up I found it difficult as guides tend to be for the SKR 1.3.

### Related Issues

None


jagoosw:
Intended to commit to bugfix so closing and will re do there

thinkyhead:
The `import-2.0.x` branch is the correct destination.

PR #161, opened 2020-07-22
Add support for the Ender 3 Pro refresh.

mindspawn:
Add support for the updated Ender 3 Pro. Seems like it is a hybrid of the Ender 3 V2 and the Ender 3 Pro.

### Requirements

* Ender 3 Pro with the Creality v4.2.2 PCB that has started to ship in the past 1-2 weeks to some folks.

### Description

With no mention on their website Creality seems to be shipping Ender 3 Pros with the STM32F SoCs.  After posting about this on Reddit seems like others have also started to receive this config (https://www.reddit.com/r/Ender3Pro/comments/hu9i9e/received_an_ender_3_pro_v2/)

Compiled and verified on my Ender 3 Pro refresh.

### Benefits

Support for very recent updates to the Ender 3 Pro refresh (new SoC).

### Related Issues

None

### DIFF
--- a/config/examples/Creality/Ender-3 Pro V1.5/Configuration.h
+++ b/config/examples/Creality/Ender-3 Pro V1.5/Configuration.h
@@ -1,0 +1,2335 @@
+/**
+ * Marlin 3D Printer Firmware
+ * Copyright (c) 2020 MarlinFirmware [https://github.com/MarlinFirmware/Marlin]
+ *
+ * Based on Sprinter and grbl.
+ * Copyright (c) 2011 Camiel Gubbels / Erik van der Zalm
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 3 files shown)


mindspawn:
The box and sticker on the unit says Ender 3 Pro. There is no v1.5 or v2 on the unit. Which is why I had to make up the v1.5 name. Its not a v2 since the v2 has a fancier DTWIN LCD panel and this unit still uses the CR-10 display.

ricsmo:
I received an Ender 3 (not Pro) from Comgrow on Amazon with this board about a week ago. I've tried to flash a bin from this and it seems to flash fine, but gets stuck on the Marlin boot screen.  Is there something Ender 3 specific to change?  I looked though the Ender 3 sample configs and changed anything in the 1.5 config that didn't look new board specific, but that didn't work either :(

Thanks. I'm totally new to this and bought the 3 for easy upgrades, etc. and they threw this at me :)

I'd also appreciate it if anyone has the original Creality .bin that they can share.

mindspawn:
Hi @ricsmo, the latest change I based my changes against were bugfix-2.0.x branch 7/18 (last change of that day).
I just tried syncing to the 7/21 changes and it did the same as what you described. Looks like someone may have broken the config. Please try building my config from 7/18 (SHA1: e85d828b24102ef4204d4792dad24e320d42cfd9) and let me know if it works for you. **_Don't forget to rename the bin file differently each time_** since as you can see in my readme this board remembers the last flashed file and ignores it in future (I wasted a lot of time due to this).
My config here is working great so far. Been printing all day :) let me know if you need any help.

ricsmo:
That worked! Haven't tried to print anything yet, but the menus loaded :)

I'll try a quick print then probably break it again setting up BLTouch :)

GijsvanDulmen:
Are these changes also for the 4.2.7 board? I ordered a silent board 1.1.5 and received the 32 bits 4.2.7 board with STM32 as well.

PR #164, opened 2020-07-23
Update FYSETC AIO example configurations

GerogeFu:
Update FYSETC AIO example configurations to support 4*TMC2208 module and 4*TMC2209 module


thinkyhead:
For completeness, please apply all possible stepper driver types to all the other configurations also. 😝

Issue #165, opened 2020-07-24
Ender 3 Pro - X Y and Z Home Bumps not Defined


flub9:
Building Marlin 2.0 using the Ender 3 Pro config files results in an error, variables X_HOME_BUMP, Y_HOME_BUMP, and Z_HOME_BUMP are never defined in the Configuration.h or Configuration_adv.h files.

I am attempting to fix this, and will submit a pull request if I can get to it. The terminal error readout is below:

```
Marlin\src\module\motion.cpp:77:71: error: 'X_HOME_BUMP_MM' was not declared in this scope
 #define XYZ_CONSTS(T, NAME, OPT) const PROGMEM XYZval<T> NAME##_P = { X_##OPT, Y_##OPT, Z_##OPT }
                                                                       ^
Marlin\src\module\motion.cpp:83:1: note: in expansion of macro 'XYZ_CONSTS'
 XYZ_CONSTS(float, home_bump_mm,   HOME_BUMP_MM);
 ^
Marlin\src\module\motion.cpp:77:80: error: 'Y_HOME_BUMP_MM' was not declared in this scope
 #define XYZ_CONSTS(T, NAME, OPT) const PROGMEM XYZval<T> NAME##_P = { X_##OPT, Y_##OPT, Z_##OPT }
                                                                                ^
Marlin\src\module\motion.cpp:83:1: note: in expansion of macro 'XYZ_CONSTS'
 XYZ_CONSTS(float, home_bump_mm,   HOME_BUMP_MM);
 ^
Marlin\src\module\motion.cpp:77:89: error: 'Z_HOME_BUMP_MM' was not declared in this scope
 #define XYZ_CONSTS(T, NAME, OPT) const PROGMEM XYZval<T> NAME##_P = { X_##OPT, Y_##OPT, Z_##OPT }
                                                                                         ^
Marlin\src\module\motion.cpp:83:1: note: in expansion of macro 'XYZ_CONSTS'
 XYZ_CONSTS(float, home_bump_mm,   HOME_BUMP_MM);
 ^
*** [.pio\build\sanguino1284p\src\src\module\motion.cpp.o] Error 1
Compiling .pio\build\sanguino1284p\src\src\module\printcounter.cpp.o
```

thisiskeithb:
Those values were removed and `HOMING_BUMP_MM` is now used in `bugfix-2.0.x` and the soon-to-be-released `2.0.6`:

[...]

PR #166, opened 2020-07-24
Add X/Y/Z_HOME_BUMP_MM

flub9:
### Description

Added definitions for X_HOME_BUMP_MM, Y_HOME_BUMP_MM, and Z_HOME_BUMP_MM, as well as HOMING_BUMP_DIVISOR to fix issues when using the examples in a Marlin build.

### Benefits

Fixes compatibility issues when using the Ender 3 Pro example profile with Marlin by adding missing values.

### Related Issues

#165 Ender 3 Pro - X Y and Z Home Bumps not Defined

### DIFF
--- a/config/examples/Creality/Ender-3 Pro/Configuration.h
+++ b/config/examples/Creality/Ender-3 Pro/Configuration.h
@@ -628,7 +628,11 @@
 //===========================================================================
 
 // @section homing
-
+// Homing hits each endstop, retracts by these distances, then does a slower bump.
+#define X_HOME_BUMP_MM 5
+#define Y_HOME_BUMP_MM 5
+#define Z_HOME_BUMP_MM 2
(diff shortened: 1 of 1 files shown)


thisiskeithb:
Those values were removed and `HOMING_BUMP_MM` is now used in `bugfix-2.0.x` and the soon-to-be-released `2.0.6`:

https://github.com/MarlinFirmware/Configurations/blob/c1a724f213effbea1e63bc4e1c3b986058e3696e/config/examples/Creality/Ender-3%20Pro/Configuration_adv.h#L634

`HOMING_BUMP_DIVISOR` is also found in `Configuration_adv.h`:

https://github.com/MarlinFirmware/Configurations/blob/c1a724f213effbea1e63bc4e1c3b986058e3696e/config/examples/Creality/Ender-3%20Pro/Configuration_adv.h#L635

You have to match the config version with the version of Marlin you're trying to compile for. See https://github.com/MarlinFirmware/Configurations/branches for a list of older branches to be used with non-bugfix Marlin.

PR #168, opened 2020-07-26
Revert SKR Mini E3 buffer sizes to Marlin defaults

thisiskeithb:
### Requirements

SKR Mini E3 V1.0, V1.2, and V2.0-based config.

### Description

Revert SKR Mini E3 V1.0, V1.2 and V2.0-based configs to Marlin defaults.

### Benefits

Some users experienced pauses before the end of prints and reverting to Marlin defaults resolved the issue.

### Related Issues

https://github.com/MarlinFirmware/Marlin/issues/18422

### DIFF
--- a/config/examples/Creality/Ender-3/BigTreeTech SKR Mini E3 1.0/Configuration_adv.h
+++ b/config/examples/Creality/Ender-3/BigTreeTech SKR Mini E3 1.0/Configuration_adv.h
@@ -1843,16 +1843,16 @@
 #if BOTH(SDSUPPORT, DIRECT_STEPPING)
   #define BLOCK_BUFFER_SIZE  8
 #elif ENABLED(SDSUPPORT)
-  #define BLOCK_BUFFER_SIZE 32
+  #define BLOCK_BUFFER_SIZE 16
 #else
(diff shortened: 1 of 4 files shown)


swilkens:
For what it's worth, I haven't noticed issues on my V1.2

thinkyhead:
Is this in reference to SD printing?

An issue has been addressed in the past, where Marlin gets to the end of the SD card and triggers some immediate response while there are still commands left in the buffer.

So, if there are mysterious pauses at the end of SD printing in Marlin now, that is the kind of thing I would look for as the primary cause.

thisiskeithb:
I’m not sure. I couldn’t reproduce the issue, but there were reports in https://github.com/MarlinFirmware/Marlin/issues/18422 that reverting these values back to stock values fixed it for users that experienced pauses.

uski:
This breaks my printer, I can no longer use the LCD screen. Changing BLOCK_BUFFER_SIZE back to 32 from 16 fixes the issue. It seems like several other users are having the same problem than I have, and this is much more serious than a "bulge" in print, so I strongly recommend reverting this... or at the very least changing BLOCK_BUFFER_SIZE to 32.

Issue #172, opened 2020-07-30
Remove Check for If HAS_SDCARD_CONNECTION


diabl0w:
https://github.com/MarlinFirmware/Configurations/blob/bb29dafbd0046a88389541203c5b98a3800a4fe0/config/examples/Creality/Ender-3/BigTreeTech%20SKR%20Mini%20E3%201.2/Configuration_adv.h#L1283

This is no longer used by Marlin, just remove the if check and keep the SDCARD_CONNECTION definition of whether it is LCD or ONBOARD

I solved this after not understanding why my SDCARD wasnt detected even after changing to ONBOARD. It was because the IF check failed so it just stayed to the default LCD value

thisiskeithb:
Thanks for the heads up! I've put in https://github.com/MarlinFirmware/Configurations/pull/174 to address this.

Issue #173, opened 2020-07-31
Stock Ender-3 config no longer builds with melzi environment


thisiskeithb:
This is mostly a reminder for myself to test all the Creality/melzi configs, but building the default Ender-3 config no longer builds with the `melzi` environment because the firmware is too big. It does work with `melzi_optimized`, but I haven't tested it on real hardware:

### `melzi`

```prolog
Error: The program size (128326 bytes) is greater than maximum allowed (126976 bytes)
RAM:   [===       ]  27.9% (used 4563 bytes from 16384 bytes)
*** [checkprogsize] Explicit exit, status 1
Flash: [==========]  101.1% (used 128326 bytes from 126976 bytes)
```

### `melzi_optimized`

```prolog
RAM:   [===       ]  27.9% (used 4563 bytes from 16384 bytes)
Flash: [========= ]  94.2% (used 122506 bytes from 130048 bytes)
```

We were already pushing the limits of these smaller boards for a while, but Marlin gained a little weight recently 😅

---

For those that find this issue, enabling `SLIM_LCD_MENUS` reduces used flash enough to fit Marlin on a stock Creality Melzi board if you are compiling with the `melzi` environment.

PR #174, opened 2020-07-31
Remove HAS_SDCARD_CONNECTION check

thisiskeithb:
### Description

Remove `HAS_SDCARD_CONNECTION` check. I searched the rest of the configs and these are the only two that still had this check.

### Benefits

`SDCARD_CONNECTION` is properly enabled.

### Related Issues

https://github.com/MarlinFirmware/Configurations/issues/172


thisiskeithb:
I haven't gone through the rest of the configs to see if they are lagging behind bugfix in other areas, but it's on my to-do list since I have to also address the default Creality/Ender-3 melzi builds no longer compiling with the latest Marlin code (https://github.com/MarlinFirmware/Configurations/issues/173).

PR #175, opened 2020-07-31
Add README.md with flashing instructions

TheNitek:
### Description

* Add instructions on how to flash the Trigorilla using a USB cable

### Benefits

* People will know how to flash

### DIFF
--- a/config/examples/delta/Anycubic/Predator/Configuration_adv.h
+++ b/config/examples/delta/Anycubic/Predator/Configuration_adv.h
@@ -1524,6 +1524,9 @@
   //
   #define FSMC_UPSCALE 3
 
+  #define LCD_FULL_PIXEL_WIDTH 480
+  #define LCD_FULL_PIXEL_HEIGHT 320
+
   //
   // Change colors
   // some colors are predefined, see /src/lcd/dogm/u8g_dev_tft_480~.cpp Line 160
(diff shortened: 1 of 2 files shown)


config/examples/delta/Anycubic/Predator/Configuration_adv.h
```diff
@@ -1524,6 +1524,9 @@
   //
   #define FSMC_UPSCALE 3
 
+  #define LCD_FULL_PIXEL_WIDTH 480
+  #define LCD_FULL_PIXEL_HEIGHT 320
```
thinkyhead: These should not be needed with the updates in MarlinFirmware/Marlin#18903.

TheNitek: The predator uses a FSMC_GRAPHICAL_TFT not the serial display Anycubic uses for the i3 (and for the Chiron?). I couldn't find anything in the referenced PR that would affect that. (Having less config would of course be nice)

thinkyhead: The values given here for `LCD_FULL_PIXEL_WIDTH`/`HEIGHT` should now come out as a consequence of `FSMC_UPSCALE 3` without these lines being needed. But I will do a test to make sure.

TheNitek: Found it scaled_tft.h is taking care of it now. Awesome!

PR #178, opened 2020-08-02
Automatically enable SDCARD_READONLY for Ender-3 and Ender-3-Pro

### LINKED ISSUE #173: Stock Ender-3 config no longer builds with melzi environment

maweil:
### Description
This PR automatically enables the SDCARD_READONLY option if no other feature like Power Loss Recovery, EEPROM Emulation or Binary File Transfer has to write to the SD card.

Please let me know what you think about this. Any feedback is welcome.

### Benefits
This saves 3214 bytes of flash storage when building the _melzi_ target without impacting the functionality.
With this change, the firmware fits into the stock motherboard again without using _melzi_optimized_.

### Related Issues
This fixes #173 for the Ender 3 and Ender 3-Pro

### DIFF
--- a/config/examples/Creality/Ender-3 Pro/Configuration_adv.h
+++ b/config/examples/Creality/Ender-3 Pro/Configuration_adv.h
@@ -1118,7 +1118,7 @@
   // Enable this option and set to HIGH if your SD cards are incorrectly detected.
   //#define SD_DETECT_STATE HIGH
 
-  #if DISABLED(POWER_LOSS_RECOVERY)
+  #if DISABLED(POWER_LOSS_RECOVERY) && DISABLED(BINARY_FILE_TRANSFER) && DISABLED(SDCARD_EEPROM_EMULATION)
     #define SDCARD_READONLY                 // Read-only SD card (to save over 2K of flash)
   #endif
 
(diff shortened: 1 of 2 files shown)


thisiskeithb:
Interesting. That's one way to handle the larger firmware.

Is there a downside to compiling with `melzi_optimized`? I can't easily run a stock Creality V1.1.x board at the moment to run test prints.

maweil:
I have not experienced any issues so far using a version compiled with `melzi_optimized` on my stock Creality Ender-3-Pro (8 bit) motherboard. I did one print without any issues after installing Marlin 2.0.6 yesterday with `SDCARD_READONLY` and some additional features enabled.

The change in this PR did still save 2992 bytes of flash when compiling with `melzi_optimized` which is then available for additional features.

PR #180, opened 2020-08-06
Update Ender-5 SKR 1.4 Turbo config + add readme

thisiskeithb:
### Description

Rename Ender-5 SKR 1.4 Turbo config to reflect that sensorless homing is enabled.

### Benefits

Sensorless homing is enabled by default in this config, which causes issues for users still running the original (real) endstops since it's not clear this feature is enabled and causes conflicts when trying to home the printer.

### Related Issues

None.


thisiskeithb:
Turns out, this has even more mods to it over a "stock" Ender-5 config that were causing several users on Facebook issues. 

From https://github.com/MarlinFirmware/Configurations/pull/159:

> - TMC2209 drivers, I will add others at a later date, probably more efficiently
> - E3D all metal hotend - changes thermistor and max temps
> - Extreme mode for all metal hotend - this sets the max temp higher which is not recommended for long term use by E3D
> - Chamber setup - configures chamber with the thermistor in the E1 thermistor pin on the SKR 1.4 and the heater in the second heater pins beside the bed heater.

~This really needs a readme with a full list of mods since users are trying to use this config after upgrading JUST their motherboard to an SKR 1.4 & using TMC2209s and not getting expected results.~ Done!

Issue #194, opened 2020-08-28
Ender 3 Pro build fails with Power Loss Recovery


pmjdebruijn:
```
In file included from Marlin/src/HAL/STM32F1/../../inc/MarlinConfig.h:41:0,
                 from Marlin/src/HAL/STM32F1/HAL_SPI.cpp:35:
Marlin/src/HAL/STM32F1/../../inc/SanityCheck.h:2183:6: error: #error "POWER_LOSS_RECOVERY is incompatible with SDCARD_READONLY."
     #error "POWER_LOSS_RECOVERY is incompatible with SDCARD_READONLY."
```

The issue occurs here, as POWER_LOSS_RECOVERY is defined after it is checked, thus NONE erroneously concludes it's okay to set SDCARD_READONLY (mind the line numbers):

https://github.com/MarlinFirmware/Configurations/blob/import-2.0.x/config/examples/Creality/Ender-3%20Pro/Configuration_adv.h#L1136

```
  #if NONE(POWER_LOSS_RECOVERY, BINARY_FILE_TRANSFER, SDCARD_EEPROM_EMULATION)
    #define SDCARD_READONLY                 // Read-only SD card (to save over 2K of flash)
  #endif

```

With this decommented (modified):
https://github.com/MarlinFirmware/Configurations/blob/import-2.0.x/config/examples/Creality/Ender-3%20Pro/Configuration_adv.h#L1167

```
  //#define POWER_LOSS_RECOVERY
```

maweil:
Thanks for analyzing this. The check in line 1136-1138 [https://github.com/MarlinFirmware/Configurations/blob/863313fca04e35430a78668cae1d7d6bcc52a1d1/config/examples/Creality/Ender-3%20Pro/Configuration_adv.h#L1136-L1138](https://github.com/MarlinFirmware/Configurations/blob/863313fca04e35430a78668cae1d7d6bcc52a1d1/config/examples/Creality/Ender-3%20Pro/Configuration_adv.h#L1136-L1138)

[...]

PR #196, opened 2020-08-30
Ender 3/Ender 3-Pro: Fix build with Power Loss Recovery enabled

maweil:
### Description

Pull request https://github.com/MarlinFirmware/Configurations/pull/178 automatically enabled `SD_CARD_READONLY` if no other features are enabled in the configuration that require write support. As described in issue #194, the check was done too early and breaks building with `POWER_LOSS_RECOVERY` or `BINARY_FILE_TRANSFER` enabled. This PR moves the check to the end of the SD support section after both features are defined.

As the original motherboards of Ender-3 (v1) and Ender-3 Pro (v1) have actual EEPROM the `SD_CARD_EEPROM_EMULATION` option is always turned off.

#### Drawbacks

When using other motherboards, automatically enabling `SD_CARD_READONLY` could still cause problems. There are multiple source files (for other motherboards) in the Marlin repository that may enable `SDCARD_EEPROM_EMULATION`. 

The following places may enable `SDCARD_EEPROM_EMULATION`:

  - [MarlinFirmware/Marlin/Marlin/src/HAL/STM32/inc/Conditionals_post.h#L26](https://github.com/MarlinFirmware/Marlin/blob/92076c65601de5fdfe0facf32dd60c57aab05747/Marlin/src/HAL/STM32/inc/Conditionals_post.h#L26)
- [MarlinFirmware/Marlin/Marlin/src/HAL/STM32F1/inc/Conditionals_post.h#L26](https://github.com/MarlinFirmware/Marlin/blob/92076c65601de5fdfe0facf32dd60c57aab05747/Marlin/src/HAL/STM32F1/inc/Conditionals_post.h#L26)
- [MarlinFirmware/Marlin/Marlin/src/pins/stm32f1/pins_CREALITY_V4.h#L56](https://github.com/MarlinFirmware/Marlin/blob/9ade72aee94981b26be0ca125ab487166d0bd7e3/Marlin/src/pins/stm32f1/pins_CREALITY_V4.h#L56)
- [MarlinFirmware/Marlin/Marlin/src/pins/stm32f1/pins_MKS_ROBIN.h#L52](https://github.com/MarlinFirmware/Marlin/blob/9ade72aee94981b26be0ca125ab487166d0bd7e3/Marlin/src/pins/stm32f1/pins_MKS_ROBIN.h#L52)

As some of these files are included later, checking for this option does not work in the configurations. If this configuration should work with other motherboards without changes to this check, reverting commit https://github.com/MarlinFirmware/Configurations/commit/d48d76739d970dd311dba4b65eb451e2d92b2a2b should be considered instead.

Any feedback is welcome.

### Benefits

This PR fixes issue #194 for Ender-3 and Ender-3 Pro.

### Related Issues
#194


thinkyhead:
The `Configuration_adv.h` file can never see `SDCARD_EEPROM_EMULATION` because it is set later.

PR #226, opened 2020-09-26
Add Ender-3 Pro Creality V4.2.7 Configs

ellensp:
### Requirements

An Ender 3 Pro machine with a Creality V4.2.7 motherboard. 
Almost identical to Ender 3 V1 with a Creality V4.2.7 added yesterday.

### Description

Add this combination to the example configurations

### Benefits

Make it easy for users to upgrade their Ender 3 Pro to this motherboard. 

### Related Issues
None


apateluk:
How does this differ to the V1.5 config ?

I'm using the Ender 3 Pro that originally came with the v4.2.2 board and updated to the v4.2.7. Should I use this instead of the V1.5 configs? 
Thanks

PR #234, opened 2020-09-30
Added CTC i3 pro B for Marlin 1.1.7 and Anet board

mytja:
### Requirements

* You'll need a CTC Prusa i3 Pro B with a Anet 1.0 board
* You'll also need to upload a bootloader

### Description

This is a self made configuration file for CTC Prusa i3 Pro B for Marlin 1.1.7 w/ Anet 1.0 board

### Benefits

It adds a support for this outdated printers (latest version from manufacturer is 1.0.1)

### Related Issues

Yes! During the testing i noticed some issues, that I hope I will fix. Can you give me a tip on how to do it?
- During start of 3d printing, Z axis moved very fast, and produced an annoying (bhrrrr) sound (like it's stuck)
- Also prints don't always start on the middle. Some are very moved to the edge, but after restarting, it's okay.
- I'm planning to add that after print is stopped, the temperature keeps its level (useful at prints, that have bed temp. more than 70 degrees celcius (Centegre)

Thank you

### DIFF
--- a/config/examples/CTC/CTC i3 Pro B/Configuration.h
+++ b/config/examples/CTC/CTC i3 Pro B/Configuration.h
@@ -1,0 +1,1785 @@
+ /*
+ * Marlin 3D Printer Firmware
+ * Copyright (C) 2016 MarlinFirmware [https://github.com/MarlinFirmware/Marlin]
+ *
+ * Based on Sprinter and grbl.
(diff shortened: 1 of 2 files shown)


thinkyhead:
Sorry, we don't have a place for these. The earliest version of Marlin that this repository supports is Marlin 1.1.9.

mytja:
Okay

PR #239, opened 2020-10-04
Fix Anet A8 Z axis stepper driver number

thearaks:
Anet A8 has a dual stepper driver / dual stepper motor configuration on the Z axis.
This PR fixes the NUM_Z_STEPPER_DRIVERS in the example configuration of the Anet A8.

### DIFF
--- a/config/examples/Anet/A8/Configuration_adv.h
+++ b/config/examples/Anet/A8/Configuration_adv.h
@@ -548,7 +548,7 @@
 //
 // For Z set the number of stepper drivers
 //
-#define NUM_Z_STEPPER_DRIVERS 1   // (1-4) Z options change based on how many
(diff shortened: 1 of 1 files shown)


thisiskeithb:
Both Z steppers are powered off a single Z driver and looking at the [Anet A8 pins file](https://github.com/MarlinFirmware/Marlin/blob/78fc4c89476bab46c6e8eecc9184f7ebb0244366/Marlin/src/pins/sanguino/pins_ANET_10.h) (`BOARD_ANET_10`) confirms that there are only four drivers.

image

thearaks:
@thisiskeithb You're right,
I probably have a modded Anet A8 right here... Sorry for the inconvenience!

PR #243, opened 2020-10-05
E current should be 770mA on Ender-3 with BigTreeTech SKR Mini E3 1.2

mtiutiu:
### Description

This fixes under-extrusion issues on the original **Ender-3** machine when using the **BTT Trinamic V1.2** board. The measured current on the **original** and **untouched** board that comes with this machine reports a value of **770mA** for the **extruder** **motor** (**0.77V** measured on **Vref** which translates to 0.77A or 770mA). The default value that is set in the configuration file is too low (**650** which corresponds to 650mA or 0.65A).

I tested on real hardware using the above mentioned board with the new setting and everything works as expected now and no motor overheating (it's only softly warm). Also the printer is using the original mechanical parts and no extra added stuff.

I haven't tested with other boards that use the same Trinamic drivers but I assume the setting should be the same as we're talking about stock parts like the extruder motor combined with Trinamic drivers so the current used to drive the motor should have the same value. I understand that maybe the default settings are lower in order to be safer but for someone that doesn't have too much knowledge of the let's say advanced settings this can cause frustration (for me it did already). The new value that I propose is safe anyways so I don't see any problem here (the extruder motor is a bigger one compared to the rest and it needs to be more capable hence the bigger current value that it needs to keep torque and to avoid missing steps).

See for yourself:
proof

### Benefits

Fixes under-extrusion and bad prints caused by the latter.

### DIFF
--- a/config/examples/Creality/Ender-3/BigTreeTech SKR Mini E3 1.2/Configuration_adv.h
+++ b/config/examples/Creality/Ender-3/BigTreeTech SKR Mini E3 1.2/Configuration_adv.h
@@ -2306,7 +2306,7 @@
   #endif
 
   #if AXIS_IS_TMC(E0)
-    #define E0_CURRENT      650
+    #define E0_CURRENT      770
     #define E0_MICROSTEPS    16
     #define E0_RSENSE         0.11
     #define E0_CHAIN_POS     -1


swilkens:
For documentation purposes, the firmware config provided by BTT for this board lists 650 mAh. We would be deviating from manufacturer config here.

https://github.com/bigtreetech/BIGTREETECH-SKR-mini-E3/blob/master/firmware/V1.2/Marlin-2.0.x-SKR-Mini-E3-V1.2/Marlin/Configuration_adv.h#L2296

But alternative sources for information on the stock ender 3 V1 board also show different values:
https://gist.github.com/knoopx/e6c40a009e796203b93a75a3ed6a5ab8

@thisiskeithb opinions?

Considering we don't have a high amount of users complaining about under-extrusion issues on v1.2 boards, I'm wondering if this is an isolated issue?

thisiskeithb:
This should be reverted back to 650 (which some users complained about being too high before, but was already tested to be OK).

It doesn’t make sense why it’s all of a sudden not OK for this user, so it shouldn’t have been changed for everyone.

swilkens:
I also seem to remember that the reported current for these motors (1A) is actually wrongly reported as peek current, not allowable RMS.

thisiskeithb:
> I also seem to remember that the reported current for these motors (1A) is actually wrongly reported as peek current, not allowable RMS.

Yes. I put in the original PRs to add these TMC current values to many of the Creality configs because people were cooking their steppers 🙂 See https://github.com/MarlinFirmware/Marlin/pull/15115 and https://github.com/MarlinFirmware/Marlin/pull/15666

I reverted the change from this PR in https://github.com/MarlinFirmware/Configurations/pull/267.

thisiskeithb:
Also, basing your changes on BigTreeTech's examples are a bad idea since their firmware has had several issues recently like only using a loosely-modified base config with incorrect settings for TMC slave addresses, default current values, etc., and several other problems that I had to put in a PR to fix/use the configs from the Marlin configs repo since I built those and have tested them on real hardware.

swilkens:
I now recall BTT had the peek power levels in their default config indeed.

mtiutiu:
Yes, BTT has it set to 800 for ALL stepper motors which is so wrong. Because of that the X, Y, Z motors started to heat - not good. So I took my original Creality board and measured Vref for each axis motor: X, Y, Z was each set to 0.65V and extruder was at 0.77V. So based on the results the current should be 650mA for X, Y, Z and 770mA for the extruder. Again I want to mention that I didn't touched the original board reference voltage for each motor driver and the stepper motors are the original ones.

So I don't understand why the extruder motor current value has to be lowered that much giving the fact that it needs more torque in order to be able to push the plastic efficiently. It's also a bigger motor physically speaking compared to the other ones.

Now of course I don't know if the Ender-3 suffered some changes lately from the manufacturer or how many hardware revisions it has gone through but still I tried to explain my reasons based on real facts and issues that I faced since I went with the BTT board and its default current settings for the TMC drivers.

I only have the BTT v1.2 so I cannot vote for the other BTT boards revisions out there but if we're talking about the same Ender-3 machine with original stock parts then the motor amperage should be the same for all. I noticed that all BTT boards use TMC2208/2209 so the values should be the same.

If this PR breaks things or if there are other aspects that I don't know of or that I missed then please feel free to undo the changes. I created it thinking that I would help others facing the same extrusion problems that I had.

Forgot to mention: I have the very first model of Ender-3 that appeared (V1 or whatever it's called).

Regards.

swilkens:
How are you arriving at 0.77V Vref = 0.77 mAh for the original board?

The manufacturer's specs may list the extruder stepper at Max power = 1A (https://ecksteinimg.de/Datasheet/Schrittmotor/JK42HS40-1004A/JK42HS40-1004AC.pdf). But the thought is (confirmed by manufacturer to @thisiskeithb ?) that this is peek allowable power, while the Marlin calculations assume RMS.

`RMS = ~1.41 * iMax -> 1.41 * 1.0 = 0.71 A RMS`

On top of this, a safety margin is taken. If we take a 10% safety margin, we would get 0.9 * 0.71 = 0.64A, or 640 mAh

mtiutiu:
@swilkens I measured it using a multimeter (UT61E).
Red probe on the cursor of the adjustable resistor and black on ground - simple as that.

Here's the extruder motor that I have:

IMG_20201014_121803

Here's the [datasheet](https://www.datasheet4u.com/mobile/977515/17HD2024N.html) that I found for it. It is specified as being rated at 1A max. As far as I know RMS calculus is usually valid for AC sine waves and not for things that are drived via pulses where duty cycle comes in. Anyways I don't want to start a debate here but I tested the 770 value already via some prints and the motor is heating only a little. I got rid of under-extrusion also so I'm happy.

thisiskeithb:
Is that an aftermarket motor? I wasn’t aware of Creality switching to Moons’.

Also, when and where did you purchase your printer?

mtiutiu:
@thisiskeithb It's the motor that I received with the printer when I bought it one year ago. I bought it from Aliexpress and was delivered from an EU warehouse (Czech republic).

thisiskeithb:
This is the first time seeing Moons’ motors on a Creality machine, so I’m wondering how many are out there.

Increasing the current for the default config still shouldn’t have occurred since most of the steppers out there are rated differently/will burn up if you run them at nearly 800 mAh.

My pending PR will revert the change, but I’ll see about adding a readme with more info on the Moons’ motor once I do some digging to find out more.

mtiutiu:
@thisiskeithb sounds great. Thanks.

swilkens:
> @swilkens I measured it using a multimeter (UT61E).
> Red probe on the cursor of the adjustable resistor and black on ground - simple as that.
> 
> Here's the extruder motor that I have:
> 
> [snip]
>  
> Here's the [datasheet](https://www.datasheet4u.com/mobile/977515/17HD2024N.html) that I found for it. It is specified as being rated at 1A max. As far as I know RMS calculus is usually valid for AC sine waves and not for things that are drived via pulses where duty cycle comes in. Anyways I don't want to start a debate here but I tested the 770 value already via some prints and the motor is heating only a little. I got rid of under-extrusion also so I'm happy.

It seems your ender 3 isn't "fully stock". Those are different motors, it makes sense that you see an improvement with that stepper model. Moon's does specify RMS, afaik.

Did you pick this printer up at saintsmart's aliexpress store or something? Perhaps a re-seller?

I hope this doesn't mean there is yet another configuration of the Ender 3 out there.

thisiskeithb:
Creality may have been forced to use Moons temporarily due to supplier issues at one point. [Here’s a post on Reddit that also shows a Moons stepper](https://reddit.com/r/ender3/comments/cr5w9r/nice_move_creality_my_order_for_an_ender_3/).

> I hope this doesn't mean there is yet another configuration of the Ender 3 out there.

I’ve looked around Creality’s various sites for firmware and haven’t seen anything specific for the Moons’ extruder version, but that’s not saying much since they aren’t great about posting firmware sources.

swilkens:
Seems there's also JK42HS40-1704XA-01F out there, which are the same manufacturer as the original motor - but a 1.7A version. 

If you look on the manufacturer's current offered models, you can also see they no longer actively offer the 1004 (1A MAX) version anymore, only a 1304 and this 1704.

Either way, it wouldn't be surprising if they go on a batch-by-batch basis for steppers and take whatever is cheapest at the time. Nightmare to maintain but they probably don't much care.

I would agree it's probably best to stay with the original batch config (650 on E) and perhaps be aware that users with missing steps on these printers may need to check the stepper model.

mtiutiu:
@swilkens @thisiskeithb Yeah..well when it comes to chinese stuff nothing is consistent from my experience but it's cheap so I cannot complain 😃.

PR #259, opened 2020-10-09
Move and rename "Ender-3 Pro V1.5" to "Ender-3 Pro/CrealityV422" for consistency

Gamester17:
Requirements

### Description

Move "Ender-3 Pro V1.5" to "Ender-3 Pro/CrealityV422" for consistency.

### Benefits

All the directories/folders under the Creality directory/folder will use the same naming standard.


ghost:
Also, why can't we just merge the Ender 3 and Ender 3 Pro Folders as well? They are basically the exact same machine with minor cosmetic changes to the hardware, but the electronics are the exact same. Also, there appear to be many differences in the configs between them as well. i.e. one use junction deviation as default and the other uses classic jerk. Ideally, they should be the exact same. It'd make more sense to provide different default configs by the board rather the machine in this case. i.e. Ender-3/422, 427, etc.

PR #263, opened 2020-10-11
Multiple fixes for example errors and warnings

sjasonsmith:
### Description

I built nearly all examples to check for errors and warnings. I have included them all in a single PR, but have left it as multiple commits so the intent of each change is more clear.

Most changes are very straighforward. A couple recurring fixes required me to make some decisions, as outlined below:
- To prevent sanity checks due to LIN_ADVANCE and S_CURVE_ACCELERATION, I have disabled S_CURVE_ACCELERATION
- To reduce size, I first disabled ARC_SUPPORT, then enabled SLIM_LCD_MENUS if needed
- If configuration was simply incomplete, I disabled features such as FILAMENT_RUNOUT_SENSOR, rather than guessing a proper configuration.

### Benefits

Reduces complaints of errors/warnings when building examples.

### Related Issues

This Marlin issue is fixed:
https://github.com/MarlinFirmware/Marlin/issues/19576

### DIFF
--- a/config/default/Configuration.h
+++ b/config/default/Configuration.h
@@ -2307,6 +2307,16 @@
  */
 //#define TFT_ROTATION TFT_NO_ROTATION
 
+/**
+ * TFT Rotation. Set to one of the following values:
+ *
+ *   TFT_ROTATE_90,  TFT_ROTATE_90_MIRROR_X,  TFT_ROTATE_90_MIRROR_Y,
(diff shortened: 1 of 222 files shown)


sjasonsmith:
This is a bit of a mess after the TFT refactor. I'll repost these after making sure they aren't doing anything they shouldn't.

PR #269, opened 2020-10-15
Update BTT SKR E3-DIP V1.x params for better experience

alfonder:
### Requirements

### Description

This config update includes the next changes:
- More default values from BigTreeTech [repo](https://github.com/bigtreetech/BIGTREETECH-SKR-E3-DIP-V1.0/tree/master/Firmware/Marlin-2.0.5_SKR-E3-DIP)
- Clean up space changes that 'diff' to the default config would generate less false positive differences.
- Speedup serial-over-USB connection for better Octoprint performance.
- Disable by default forgotten BLtouch support.
- Increase too short (annoying) timeout of auto return to Status screen.
- Enable Babystepping.
- Set Linear Advance K=0 to avoid known problems with TMC2208.

### Benefits

Config may be used immediately without (or with minor) modifications with the latest Marlin version for Creality Ender-3 with a popular BigTreeTech SKR E3-DIP + TMC2208 motherboard upgrade.

### Related Issues

None.

### DIFF
--- a/config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration.h
+++ b/config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration.h
@@ -120,7 +120,7 @@
  *
  * :[2400, 9600, 19200, 38400, 57600, 115200, 250000, 500000, 1000000]
  */
-#define BAUDRATE 115200
+#define BAUDRATE 250000
 
 // Enable the Bluetooth serial interface on AT90USB devices
 //#define BLUETOOTH
@@ -131,7 +131,7 @@
 #endif
 
 // Name displayed in the LCD "Ready" message and Info menu
-#define CUSTOM_MACHINE_NAME "Ender-3"
+#define CUSTOM_MACHINE_NAME "Ender-3 SKR"
 
 // Printer's unique ID, used by some programs to differentiate between machines.
 // Choose your own or use a service like https://www.uuidgenerator.net/version4
@@ -473,7 +473,7 @@
 #define HEATER_5_MAXTEMP 275
 #define HEATER_6_MAXTEMP 275
 #define HEATER_7_MAXTEMP 275
-#define BED_MAXTEMP      125
+#define BED_MAXTEMP      150
 
 //===========================================================================
 //============================= PID Settings ================================
@@ -492,18 +492,16 @@
   //#define PID_PARAMS_PER_HOTEND // Uses separate PID parameters for each extruder (useful for mismatched extruders)
                                   // Set/get with gcode: M301 E[extruder number, 0-2]
 
-  // If you are using a pre-configured hotend then you can use one of the value sets by uncommenting it
-  // Creality Ender-3
   #if ENABLED(PID_PARAMS_PER_HOTEND)
     // Specify between 1 and HOTENDS values per array.
     // If fewer than EXTRUDER values are provided, the last element will be repeated.
-    #define DEFAULT_Kp_LIST {  24.70,  24.70 }
-    #define DEFAULT_Ki_LIST {   2.55,   2.55 }
-    #define DEFAULT_Kd_LIST {  59.91,  59.91 }
+    #define DEFAULT_Kp_LIST {  22.20,  22.20 }
+    #define DEFAULT_Ki_LIST {   1.08,   1.08 }
+    #define DEFAULT_Kd_LIST { 114.00, 114.00 }
   #else
-    #define DEFAULT_Kp  24.70
-    #define DEFAULT_Ki   2.55
-    #define DEFAULT_Kd  59.91
+    #define DEFAULT_Kp  21.73
+    #define DEFAULT_Ki   1.54
+    #define DEFAULT_Kd  76.55
   #endif
 #endif // PIDTEMP
 
@@ -540,20 +538,13 @@
   //#define MIN_BED_POWER 0
   //#define PID_BED_DEBUG // Sends debug data to the serial port.
 
-  //120V 250W silicone heater into 4mm borosilicate (MendelMax 1.5+)
-  //from FOPDT model - kp=.39 Tp=405 Tdead=66, Tc set to 79.2, aggressive factor of .15 (vs .1, 1, 10)
-
-  //120V 250W silicone heater into 4mm borosilicate (MendelMax 1.5+)
-  //from pidautotune
-  //#define DEFAULT_bedKp 97.1
-  //#define DEFAULT_bedKi 1.41
-  //#define DEFAULT_bedKd 1675.16
-
-  // FIND YOUR OWN: "M303 E-1 C8 S90" to run autotune on the bed at 90 degreesC for 8 cycles.
+  // 120V 250W silicone heater into 4mm borosilicate (MendelMax 1.5+)
+  // from FOPDT model - kp=.39 Tp=405 Tdead=66, Tc set to 79.2, aggressive factor of .15 (vs .1, 1, 10)
   #define DEFAULT_bedKp 50.71
   #define DEFAULT_bedKi 9.88
   #define DEFAULT_bedKd 173.43
 
+  // FIND YOUR OWN: "M303 E-1 C8 S90" to run autotune on the bed at 90 degreesC for 8 cycles.
 #endif // PIDTEMPBED
 
 #if EITHER(PIDTEMP, PIDTEMPBED)
@@ -706,7 +697,7 @@
 
 // Enable this feature if all enabled endstop pins are interrupt-capable.
 // This will remove the need to poll the interrupt pins, saving many CPU cycles.
-//#define ENDSTOP_INTERRUPTS_FEATURE
+#define ENDSTOP_INTERRUPTS_FEATURE
 
 /**
  * Endstop Noise Threshold
@@ -723,7 +714,7 @@
 //#define ENDSTOP_NOISE_THRESHOLD 2
 
 // Check for stuck or disconnected endstops during homing moves.
-//#define DETECT_BROKEN_ENDSTOP
+#define DETECT_BROKEN_ENDSTOP
 
 //=============================================================================
 //============================== Movement Settings ============================
@@ -757,7 +748,7 @@
  * Override with M203
  *                                      X, Y, Z, E0 [, E1[, E2...]]
  */
-#define DEFAULT_MAX_FEEDRATE          { 500, 500, 5, 25 }
+#define DEFAULT_MAX_FEEDRATE          { 500, 500, 5, 50 }
 
 //#define LIMITED_MAX_FR_EDITING        // Limit edit via M203 or LCD to DEFAULT_MAX_FEEDRATE * 2
 #if ENABLED(LIMITED_MAX_FR_EDITING)
@@ -1090,7 +1081,7 @@
 // @section extruder
 
 #define DISABLE_E false             // Disable the extruder when not stepping
-//#define DISABLE_INACTIVE_EXTRUDER // Keep only the active extruder enabled
+//#define DISABLE_INACTIVE_EXTRUDER   // Keep only the active extruder enabled
 
 // @section machine
 
@@ -1245,7 +1236,7 @@
 //#define AUTO_BED_LEVELING_LINEAR
 //#define AUTO_BED_LEVELING_BILINEAR
 //#define AUTO_BED_LEVELING_UBL
-#define MESH_BED_LEVELING
+//#define MESH_BED_LEVELING
 
 /**
  * Normally G28 leaves leveling disabled on completion. Enable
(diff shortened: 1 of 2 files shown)


thisiskeithb:
Using the values from the BigTreeTech repo is not a good choice here. I've already had to put in a PR there to fix a bunch of their settings since they messed up the recent configs using lightly-modded default configs.

alfonder:
> Using the values from the BigTreeTech repo is not a good choice here. I've already had to put in a PR there to fix a bunch of their settings since they messed up the recent configs using lightly-modded default configs.

OK, at least some values helped to solve a number of problems. So, what values did you consider to alter? Let's co-operate and tune them together.

config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration.h
```diff
@@ -1645,7 +1636,7 @@
  *
  * View the current statistics with M78.
  */
-//#define PRINTCOUNTER
+#define PRINTCOUNTER
```
sjasonsmith, line 1639: I highly discourage the use of PRINTCOUNTER with anything emulating eeprom in flash. It is slow and is likely undesirable during a print. It also reduces the life of the flash chip, which is only rated for 10,000 erase cycles. Just over a year of continuous printing could render it useless.

config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration_adv.h
```diff
@@ -2487,7 +2487,7 @@
    * M912 - Clear stepper driver overtemperature pre-warn condition flag.
    * M122 - Report driver parameters (Requires TMC_DEBUG)
    */
-  //#define MONITOR_DRIVER_STATUS
+  #define MONITOR_DRIVER_STATUS
```
sjasonsmith, line 2490: This board used SoftwareSerial which may cause performance issues during prints.

config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration_adv.h
```diff
@@ -2578,7 +2578,7 @@
    * Beta feature!
    * Create a 50/50 square wave step pulse optimal for stepper drivers.
    */
-  #define SQUARE_WAVE_STEPPING
+  //#define SQUARE_WAVE_STEPPING
```
sjasonsmith, line 2581: This should certainly be left enabled.
It isn’t really noticeable at the moment, but I think it should be the default where possible. It simple does nothing if the drivers don’t support it.

config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration.h
```diff
@@ -473,7 +473,7 @@
 #define HEATER_5_MAXTEMP 275
 #define HEATER_6_MAXTEMP 275
 #define HEATER_7_MAXTEMP 275
-#define BED_MAXTEMP      125
+#define BED_MAXTEMP      150
```
sjasonsmith, line 476: Really? Do these stock beds go to 150?

sjasonsmith (commented):
I think it would be really helpful to create a PR specifically for things that are broken or incorrect. Right now this is a mix of “bug fixes” and a lot of additional changes which are simply your preference.

config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration.h
thisiskeithb: Nope.

alfonder: This is the thing I called "false positive difference". The default config has the same value:
`#define BED_MAXTEMP      150`
AFAIK this parameter defines only upper margin for Bed Temperature and does not affect print quality directly. IMHO it just doesn't need to be changed.

config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration_adv.h
alfonder: It was disabled by default and has "beta" status. Am I wrong that this parameter is meaningful only for SpreadCycle mode? And by default the extruder driver mode is stealthChop. Anyway I don't object to the option enabling.

config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration.h
thisiskeithb: This should be reverted. These beds cannot go that high and will cause your printer to halt if you try.

thisiskeithb:
> I think it would be really helpful to create a PR specifically for things that are broken or incorrect. Right now this is a mix of “bug fixes” and a lot of additional changes which are simply your preference.

I agree. There also seems to be some misunderstanding of what these changes do.

Please try to match any changes to the existing SKR Mini E3 configs hosted here since they've already been tested on real hardware and are based off the stock Ender-3 configs.

config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration_adv.h
alfonder: As I could find in reviews, this option decreases performance on 8-bit controllers only. On the other hand this parameter can help to avoid a number of problems.
Did you see any evidence of performance issues on STM32?

config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration.h
alfonder: Agreed. I will roll back this change.

alfonder:
@thisiskeithb what changes are considered as non-doubtful fixes? I'll separate them to a new PR.
Please, review the list:
- Clean up space changes that 'diff' to the default config would generate less false positive differences.
- Disable by default forgotten BLtouch support.
- Set Linear Advance K=0 to avoid known problems with TMC2208.

Is it OK?
Maybe also:
- Speedup serial-over-USB connection for better Octoprint performance.
- Increase too short (annoying) timeout of auto return to Status screen.

?

thisiskeithb:
> Clean up space changes that 'diff' to the default config would generate less false positive differences.

Which spaces?

> Disable by default forgotten BLtouch support.

What do you mean? This is already disabled.

> Set Linear Advance K=0 to avoid known problems with TMC2208.

Yes.

> Speedup serial-over-USB connection for better Octoprint performance.

This was reverted in the other SKR configs due to slow down/stuttering issues some users were experiencing, so it shouldn't be enabled for everyone.

> Increase too short (annoying) timeout of auto return to Status screen.

That's a personal preference, so I don't think it should be pushed to the config.

Again, please see the other SKR Mini E3 config as an example as they should be as closely synced as possible.

config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration.h
```diff
@@ -1090,7 +1081,7 @@
 // @section extruder
 
 #define DISABLE_E false             // Disable the extruder when not stepping
-//#define DISABLE_INACTIVE_EXTRUDER // Keep only the active extruder enabled
+//#define DISABLE_INACTIVE_EXTRUDER   // Keep only the active extruder enabled
```
alfonder, line 1084: This was space-only change, comparing with the default config.

config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration.h
```diff
@@ -1813,7 +1804,7 @@
 //
 // Add individual axis homing items (Home X, Home Y, and Home Z) to the LCD menu.
 //
-// #define INDIVIDUAL_AXIS_HOMING_MENU
+//#define INDIVIDUAL_AXIS_HOMING_MENU
```
alfonder, line 1807: This was space-only change, comparing with the default config.

config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration.h
```diff
@@ -540,20 +538,13 @@
...
-  //#define DEFAULT_bedKd 1675.16
-
-  // FIND YOUR OWN: "M303 E-1 C8 S90" to run autotune on the bed at 90 degreesC for 8 cycles.
+  // 120V 250W silicone heater into 4mm borosilicate (MendelMax 1.5+)
+  // from FOPDT model - kp=.39 Tp=405 Tdead=66, Tc set to 79.2, aggressive factor of .15 (vs .1, 1, 10)
   #define DEFAULT_bedKp 50.71
   #define DEFAULT_bedKi 9.88
   #define DEFAULT_bedKd 173.43
 
+  // FIND YOUR OWN: "M303 E-1 C8 S90" to run autotune on the bed at 90 degreesC for 8 cycles.
```
alfonder, line 547: This was comments-only change, comparing with the default config.

config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration.h
```diff
@@ -1245,7 +1236,7 @@
 //#define AUTO_BED_LEVELING_LINEAR
 //#define AUTO_BED_LEVELING_BILINEAR
 //#define AUTO_BED_LEVELING_UBL
-#define MESH_BED_LEVELING
+//#define MESH_BED_LEVELING
```
alfonder, line 1239: This is for restoring default value, because the parameter has no meaning without BLtouch (forgotten).

config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration.h
```diff
@@ -1290,7 +1281,7 @@
 #if EITHER(AUTO_BED_LEVELING_LINEAR, AUTO_BED_LEVELING_BILINEAR)
 
   // Set the number of grid points per dimension.
-  #define GRID_MAX_POINTS_X 5
+  #define GRID_MAX_POINTS_X 3
```
alfonder, line 1284: This is for restoring default value, because the parameter has no meaning without BLtouch (forgotten).

config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration.h
```diff
@@ -1300,7 +1291,7 @@
 
     // Beyond the probed grid, continue the implied tilt?
     // Default is to maintain the height of the nearest edge.
-    #define EXTRAPOLATE_BEYOND_GRID
+    //#define EXTRAPOLATE_BEYOND_GRID
```
alfonder, line 1294: This is for restoring default value, because the parameter has no meaning without BLtouch (forgotten).

config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration.h
```diff
@@ -1350,12 +1341,12 @@
...
  * Include a guided procedure if manual probing is enabled.
  */
-#define LCD_BED_LEVELING
+//#define LCD_BED_LEVELING
 
 #if ENABLED(LCD_BED_LEVELING)
   #define MESH_EDIT_Z_STEP  0.025 // (mm) Step size while manually probing Z axis.
   #define LCD_PROBE_Z_RANGE 4     // (mm) Z Range centered on Z_MIN_POS for LCD Z adjustment
-  #define MESH_EDIT_MENU        // Add a menu to edit mesh points
+  //#define MESH_EDIT_MENU        // Add a menu to edit mesh points
```
alfonder, line 1349: This is for restoring default value, because the parameter has no meaning without BLtouch (forgotten).

config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration_adv.h
```diff
@@ -708,7 +708,7 @@
    * differs, a mode set eeprom write will be completed at initialization.
    * Use the option below to force an eeprom write to a V3.1 probe regardless.
    */
-  #define BLTOUCH_SET_5V_MODE
+  //#define BLTOUCH_SET_5V_MODE
```
alfonder, line 711: This is for restoring default value, because the parameter has no meaning without BLtouch (forgotten).

alfonder:
> Again, please see the other SKR Mini E3 config as an example as they should be as closely synced as possible.

I did it first. But I got under-extrusion problem, while BTT config didn't have such problem. Therefore I borrowed a number of parameters from BTT config. I agree that some parameters should better be taken form E3 mini's config. And I asked for help. But it's not a good approach to take all from E3 mini by default, because of under-extrusion problem. 
So my aim is to make E3-DIP config work.
I plan to separate this PR: porting mistakes, such as space/comment  changes and forgotten meaningless BLtouch params - in one PR (doesn't affect resulting firmware), Linear Advance K error - in another PR, meaningful params to fix extruding - in the third PR. And the latter need to be discussed in details.

alfonder:
The PR has been separated into isolated units:
- #277 
- #278
- #279

PR #270, opened 2020-10-15
Use named TFT option for Tevo Nereus config (MKS_ROBIN_TFT35)

thisiskeithb:
### Description

This likely wasn't caught in all the TFT refactoring and there are probably some more configs that could benefit from this change, but I was helping someone out with this specific config & noticed it.

### Benefits

Easier/simplified config of the TFT. Tested on a Robin Nano 1.2  with 480x320 TFT35 (same combo used on the Nereus).

### Related Issues

None.


thinkyhead:
These will have to be resolved case by case, but there was a comment laying out the most likely candidates for some of these. I'll see if I can dig it up….

thisiskeithb:
@thinkyhead The FLSUN Q5 has a 3.2” TFT, so it would be `MKS_ROBIN_TFT32`.

See https://github.com/MarlinFirmware/Marlin/issues/19834

PR #277, opened 2020-10-19
Cleanup BTT SKR E3-DIP V1.x config

alfonder:
### Requirements

### Description

This patch set brings the following config parts according to the default config:
- Whitespace and comment differences
- Enabling/setting meaningless parameters (BLtouch- and LCD-related)
- Renamed parameters

### Benefits

The main goal is that comparing the default config with the board config would not show hunks, which are actually changes nothing. For example, for command:
`diff -u config/default/Configuration.h "config/examples/Creality/Ender-3/BigTreeTech SKR E3-DIP V1.1/Configuration.h"`
Applying this patch set will not change the resulting firmware image.

### Related Issues

This PR is a part of separating #269 into logically different units.


thinkyhead:
Good catches, thanks!

PR #283, opened 2020-10-23
Stepper motors for BQ Hephestos 2: Update Configuration.h

Bergerac56:
This is my first pull request. I hope I did not do any mistakes in the process ;)

### Requirements

### Description

With the proposed configuration files for Hephestos 2, and at least after commit ea80b2c, steppers motors are not working as expected.
On my BQ_ZUM_MEGA_3D, it seems that stepper drivers are DRV8825.

Solution:
X,Y,Z and E0 have to be adapted accordingly.

### Benefits

Solves the erratic movements of X,Y,X and E stepper motors


Bergerac56:
@thinkyhead Dear Scott, I do not see the modification of the drivers for Hephestos 2. As you closed the PR I supposed it was OK
As it is my first PR (Most of the time, I suggest and other do if relevant ;) ) was it a proble with it? Wrongly done? Not correctly proposed?

So I learn :)

Thank you

Olivier

PR #287, opened 2020-10-25
Disable S_CURVE_ACCELERATION to avoid LIN_ADVANCE sanity check

sjasonsmith:
### Description

Two examples enable both S_CURVE_ACCELERATION and LIN_ADVANCE. This breaks the builds due to a sanity check.
I have disabled S_CURVE_ACCELERATION to avoid this.

### Benefits

Examples build.

### Related Issues

N/A

### DIFF
--- a/config/examples/Creality/Ender-5 Plus/Configuration.h
+++ b/config/examples/Creality/Ender-5 Plus/Configuration.h
@@ -837,7 +837,7 @@
  *
  * See https://github.com/synthetos/TinyG/wiki/Jerk-Controlled-Motion-Explained
  */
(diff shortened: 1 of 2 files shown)


thinkyhead:
Maybe we should _enable_ `EXPERIMENTAL_SCURVE` instead…?

sjasonsmith:
I wasn't sure, I had to either disable one of the conflicting settings, or enable `EXPERIMENTAL_SCURVE`. It doesn't really matter which way it goes to me.

If our stance on `EXPERIMENTAL_SCURVE` is that it is good enough for examples, then maybe it should just go away and become the new normal?

PR #291, opened 2020-10-25
Add missing MKS SBase DigiPot addresses

sjasonsmith:
### Description

SBase example failed to build because it was missing the Digipot address.
I did not test these, but selected the Smoothieboard values since I believe the SBase runs Smoothieware with no special configuration needed.

### Benefits

Examples build.

### Related Issues

N/A


Bergerac56:
I confirm this addition. This is what I use on my MKS-SBASE for a long time now.

PR #293, opened 2020-10-25
Fix park location for Dagoma/Disco Ultimate

sjasonsmith:
### Description

Park location was unreachable, causing build error.

### Benefits

Fix build.

### Related Issues

N/A


sjasonsmith:
This should be fixed by the changes @0r31 made in #300.

PR #294, opened 2020-10-25
Disable features on sanguino1284p examples that don't fit in flash

sjasonsmith:
### Description

Two `sanguino1284p` examples are too large to build. Disable features to make them fit. I only did this for boards which only list `sanguino1284p` as a possible environment.

I did not update examples which use boards specifying `melzi` as a possible environment, since for those boards users can switch to Optiboot and use the `melzi_optimized` environment.

### Benefits

Examples builds for `sanguino1284p`.

### Related Issues

N/A


thinkyhead:
Marlin is probably overdue for size optimization in consideration of 128K boards, as the default build has been growing in size.

PR #296, opened 2020-10-25
Add important SD card information for Ender 5 plus DGUS Display flashing.

ellensp:
### Requirements

 DGUS Touch Display with DWIN OS

### Description

If you attempt to upgrade the DGUS Displays with the data from the DWIN_SET folder and the SD card is not formatted correctly It will not work. 
It displays that 0 files are loaded.
eg
DWIN Error
But does not tell you why. 
Incorrect formatting of the SD card is why.

### Benefits

Users who read the note in the Config will be able to update the DGUS Display.


ellensp:
Your "tweak"  created two step 4's
```
//  4. Plug the microSD card into the back of the display.
//  4. Boot the display and wait for the update to complete.
```

PR #302, opened 2020-10-30
Ender 5 BTT SKR E3 1.2  example

chepo92:
### Description

Ender 5 BTT SKR E3 1.2  example
Bed is configured to 230x230 as the printer can't reach the 235x235 bed


rippiedoos:
I looked at this config, but the bed should be 220x220x300, not 235x235x250 (or 230x230x250 as it is now). That's more the size of a CR-6 SE.

And is there a reason why on the SKR Mini E3 2.0 the option S_CURVE_ACCELERATION is enabled and on the E3 1.2 not? I thought the differences between both board are minimal so features like these should differ greatly.

chepo92:
Ender 5 has a measured 235x235 bed, but mechanically it can't reach bed size but 230, it hits the Y carriages and looses steps (tested with a sample of 3 ender 5), If you have an ender 5 to counter proof that maybe we should modify it. 

I think you mean S curve is enabled in this config, but disabled in SKR 2.0, IMHO S curve acc is a desired default function, and should be enabled (suggested). 
As always I'm open to talk about it on discord :D

rippiedoos:
Sorry about the mixup with the S curve acceleration.

I'm currently printing with the 230x230 setting and if it's done printing, it returns to 0,0 (x/y). And it hits the corner brackets on my ender 5. And also in the other Creatility Configs, (the V1 and V4.2.7) it has the size of 220x220x300. I think it's just weird that not all 5 current configs have the same bed size settings. And even the official Creality Ender 5 website reports the build volume as 220x220x300. 

But true, the bed is measured 235x235 but the default end stop settings prevent it from reaching the corner of the bed. It's like a centimeter off the corner of the bed if it's hitting the end stops. And yes, you could set the bed size to 235x235 but then you would need to adjust the {X,Y}_{MIN,MAX}_POS as well, to adjust for this offsets as well. 220x220 works better for me than 230x230, because it's not hitting the end stops any more. I just measured and 230x230 would be _just_ hitting the corner brackets so 225x225 would be better suited for countering a slight overshoot and not hitting the corner brackets.

My point being, I'm (maybe falsely) expecting basic settings for this printer, like bed size settings, are all the same in the 5 configs. That's not the case at the moment. And because my Ender 5 is hitting my corner brackets when it's moving to 0,0 after it has homed to 230,230, I thought I would chime in and report it that the bed size settings aren't correct or consistent across the different Ender 5 configs.

PR #324, opened 2020-11-11
Anet E16 BTT SKR13

chepo92:
### Requirements

### Description
Adds config example for an Anet E16 with board BTT SKR 1.3, with TMC2208 drivers, current for Z axis is increased from default for the two Z motors, main features enabled: S curve, mesh bed leveling
Notice: Bed is 300x300 but actually printer only can reach 280x300 

### Benefits
An example for a BTT SKR 1.3 as there is no example (that i could find in this repo) for any machine with this board


sjasonsmith:
Does this example work with stock hardware with only the motherboard replaced, or does it include additional non-stock modifications as well?

chepo92:
Right! I had the steps/mm for BMG, now fixed that

PR #330, opened 2020-11-17
Revert "Fix minor correction for Delta QQS-Pro HISPEED board. "

Foxies-CSTL:
Reverts MarlinFirmware/Configurations#309

Hello,
The previous configuration requires some adjustments.
Reading various PRs and Deltabots documentation, I noticed that the segment length calculation method was important and dependent on the processor used. I have performed several printing tests with the same GCode (Test Benchy) at different speeds: I get better results with a segment at 80 per second than with a segment at 200.
Maybe Marlin's calculation method is different from Repetier's (original firmware).
I have also made adjustments to the automatic calibration (Delta) and leveling (UBL) method so that the user is minimal adjustments to make.
Optional not activated in the configuration:
I validated the use of the TMC2209 in UART serial mode in single-wire, with an ESP-type WiFi module and management of the led strips with a modified NeoPixels library.
TMC2209
image

**The 2 modified configuration files**:
[QQS-Pro.zip](https://github.com/MarlinFirmware/Configurations/files/5585424/QQS-Pro.zip)

That's it for this time ;-)


thinkyhead:
These are pretty radical changes, and some seem automated with no thought. We'll need more review on this before it goes forward.

Foxies-CSTL:
> These are pretty radical changes, and some seem automated with no thought. We'll need more review on this before it goes forward.

These automatisms give me a lot of thoughts knowing that I work with a single machine without documentation from the manufacturer and with feedback from some users ;-)
But it's nice !!

thinkyhead:
Thanks for taking the time to do the extra review. Did you ultimately determine that these changes are not needed for most users?

Foxies-CSTL:
> Did you ultimately determine that these changes are not needed for most users?

You're welcome Scott, but it was not my intention to change everything, only to make it more compatible with all the latest changes to the Marlin code and to inform users of the capabilities introduced on this clone of the MKS_Robin_mini board.
Good luck for the future ;-)

PR #351, opened 2020-11-30
Fix touch calibration values for classic ui

rhapsodyv:
Now that classic don't have fixed touch coordinates at 320x240 and uses fully screen resolution, we need to fix those old calibration values. This are the new calibration values for this board.


sjasonsmith:
@twagler you were the last person to modify this Predator config. Are you interested in verifying these changes prior to merging this?

twagler:
Yeah, I’ll take a look at it first thing in the morning.

PR #352, opened 2020-12-01
Sapphire plus

wjohnsaunders:
Fix some issues with the Sapphire Plus configuration. The SAPPHIRE_PLUS_MKS_UI configuration item enabled the wrong menu system. In one place BLTOUCH was used before defined instead of SAPPHIRE_PLUS_BLTOUCH.

No need to differentiate max extrude length and filament load/unload distances. Starting at the max required, users can tune down via the menu if they shorten their bowden tube.

Provided tuned PID values for hotend and bed.

Enable tramming wizard menu item.

BLTOUCH needs 5mm of clearance for homing, so X_MAX_POS set to 335. Otherwise a crash will occur when homed from maximum Z height.

Added TMC UART configuration (although simple HW mods are required to make it work).


SpaceEggs24:
@wjohnsaunders Thank you for your work regarding the Sapphire configurations. I want to install Marlin using your configuration files but i have problems customizing the configuration files to my needs. 
My objective is to use the two Z-axis endstops in combination with the BL-Touch. The BL-Touch should perform a mesh probing sequenze save that and when i want to start printing i want the printer to only use the the Z endstops for homing and the use the before determined mesh to create a perfect first layer. The probe should be connected to the 2. filament runout pins on the board. 

I already have this configuration running on my Sapphire Plus but the precompiled firmware is buggy, i dont have any way of configuring it and i dont have the configuration files for reference. 

I think this configuration would be a great one for a lot of people with this printer (at least i hope). I'm not to experienced with configuring Marlin so it would be really awesome if you could look into it. Of course i will continue doing my reseach and if i am able to create my own working configuration i will provide it to anyone whos interested.

Thanks!

PR #357, opened 2020-12-03
Sapphire Pro TMC2209/BLTOUCH Easy Switch

ils15:
Enabled same Options as BlueR to Sapphire Pro

- Enable `SPRO_TMC2209` if your board has TMC2209 drivers.
- Enable `SPRO_BLTOUCH` if you want to enable BLTOUCH. (BLTOUCH Will Replace Z_Min_Plug)


config/examples/Two Trees/Sapphire Pro/Configuration.h
```diff
@@ -1082,15 +1098,23 @@
...
 #ifdef SPRO_INVERTED_E
-  #define INVERT_E0_DIR true
+  #if ENABLED(SPRO_TMC2209)
+    #define INVERT_E0_DIR false
+  #else
+    #define INVERT_E0_DIR true
+  #endif
 #else
-  #define INVERT_E0_DIR false
+  #define INVERT_E0_DIR true
```
jimmywan: Are you sure that you want to flip this value? Seems like this should remain false.

ils15: you are right

ils15 (commented):
change direction

PR #373, opened 2020-12-24
Fix more ET4 configs, unify

thisiskeithb:
### Description

I walked someone through updating a Labsists ET4 (with capacitive/non-removable probe) and found some tweaks could be made to all the ET series' instructions as well as fix some issues with the Labists ET4 config. I also removed the old `MKS_ROBIN_TFT32` TFT defines from the Anet ET4 configs.

See my comment below about the other config updates.

### Benefits

Clearer documentation, fix Labists ET4 & general ET4/5 config issues.

### Related Issues

None. This was done over Twitter/email.

### DIFF
--- a/config/examples/Anet/ET4-Pro/Configuration.h
+++ b/config/examples/Anet/ET4-Pro/Configuration.h
@@ -2490,7 +2490,7 @@
   #define BUTTON_DELAY_EDIT  50 // (ms) Button repeat delay for edit screens
   #define BUTTON_DELAY_MENU 250 // (ms) Button repeat delay for menus
 
-  //#define TOUCH_SCREEN_CALIBRATION
+  #define TOUCH_SCREEN_CALIBRATION
 
   //#define TOUCH_CALIBRATION_X -11838
   //#define TOUCH_CALIBRATION_Y   8776
--- a/config/examples/Anet/ET4-Pro/Configuration.h
+++ b/config/examples/Anet/ET4-Pro/Configuration.h
@@ -480,21 +480,27 @@
 #define PID_K1 0.95      // Smoothing factor within any PID loop
 
 #if ENABLED(PIDTEMP)
-  #define PID_EDIT_MENU         // Add PID editing to the "Advanced Settings" menu. (~700 bytes of PROGMEM)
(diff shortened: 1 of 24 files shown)


thisiskeithb:
I just tried compiling the other ET4 configs and ran into some issues with homing direction/homing with `Z_MIN_PROBE_USES_Z_MIN_ENDSTOP_PIN`.

The original ET4 homed to Z max and you ran ABL with a [removable probe](https://www.anet3d.com/product/5pcs-auto-leveling-sensor-of-et4/) (similar to a delta), but I've asked Anet to confirm if this is the only model that ever shipped with this configuration.

sjasonsmith:
Do you have questions about this probing setup? That’s similar to how a lot of deltas work.

config/examples/Anet/ET4-Pro/Configuration.h
```diff
@@ -2345,7 +2345,7 @@
 // 320x240, 3.2", FSMC Display From MKS
 // Normally used in MKS Robin Nano V1.2
 //
-#define MKS_ROBIN_TFT32
+//#define MKS_ROBIN_TFT32
```
rhapsodyv, line 2354: You should not add ANET_ET4_TFT28 or ANET_ET5_TFT35 too?

thisiskeithb: It’s already enabled.

thisiskeithb:
> Do you have questions about this probing setup? That’s similar to how a lot of deltas work.

I don’t. I just need Anet to confirm that the original ET4 is the only model with this homing/probing setup so I can fix the configs.

There’s also a mix of bilinear and UBL enabled from when Scott imported them and other minor differences, so I want to unify them even more.

thisiskeithb:
I cleaned up & unified ET4 configs since settings were all over the place. These are all the same base hardware except for:

- Original ET4 (non-plus version) homes to Z max and has a clip on probe for ABL.
- All other ET4/5 models have a capacitive probe built into the X carriage and home to Z min.
- ET4/5 Pro are the only versions with TMC2208s.

I also compiled each config individually and all compile.

PR #377, opened 2020-12-26
Update Alfawise U20 samples

tpruvot:
Sorry, made the PR on the wrong branch first...


tpruvot:
PR #370 on the right branch...

sjasonsmith:
Do you think the default should be reduced for all printers?

I saw the comment that this was just increased on 2.0.6.1. It would be good to go back and see who did it and the reason why.

tpruvot:
yes i guess, its not really visible in all situations, but a bit fast imo...

rmm, need to rebase that.. also

PR #384, opened 2020-12-31
Fix CR-10 V2 with BLTouch

Hogesyx:
Allow Z_MIN_ENDSTOP to retain functionality after enabling CR10V2_BLTOUCH

### Requirements

### DIFF
--- a/config/examples/Creality/CR-10 V2/Configuration.h
+++ b/config/examples/Creality/CR-10 V2/Configuration.h
@@ -658,10 +658,10 @@
 // Mechanical endstop with COM to ground and NC to Signal uses "false" here (most common setup).
 #define X_MIN_ENDSTOP_INVERTING false // Set to true to invert the logic of the endstop.
 #define Y_MIN_ENDSTOP_INVERTING false // Set to true to invert the logic of the endstop.
-#define Z_MIN_ENDSTOP_INVERTING ENABLED(CR10V2_BLTOUCH) // Set to true to invert the logic of the endstop.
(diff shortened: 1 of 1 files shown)


config/examples/Creality/CR-10 V2/Configuration.h
```diff
@@ -845,11 +845,13 @@
...
  */
-#define Z_MIN_PROBE_USES_Z_MIN_ENDSTOP_PIN
+//#define Z_MIN_PROBE_USES_Z_MIN_ENDSTOP_PIN
 
 // Force the use of the probe for Z-axis homing
 //#define USE_PROBE_FOR_Z_HOMING
-
+#if ENABLED(CR10V2_BLTOUCH)
+  #define USE_PROBE_FOR_Z_HOMING
+#endif
```
sjasonsmith, line 854: Won't you need `Z_MIN_PROBE_USES_Z_MIN_ENDSTOP_PIN` enabled if `CR10V2_BLTOUCH` is not defined?

Hogesyx: As far as I understand the purpose of `Z_MIN_PROBE_USES_Z_MIN_ENDSTOP_PIN ` is to use the `Z_STOP_PIN` instead of the `Z_MIN_PROBE_PIN`. The original code define this and yet did a override of the Z_STOP_PIN with the pin number of probe pin on the Creality board which is redundant steps and yet with the side effect of disabling the endstop.

sjasonsmith: You are right. I wasn’t thinking straight with my earlier comment! I was thinking it would be needed for an endstop but that is obviously not true.

PR #388, opened 2021-01-04
Add missing precompiler directives

alexlanghart:
### Description

This fixes the build error described here: https://gitmemory.com/issue/MarlinFirmware/Marlin/20623/753316550

Summarized as: 
`sketch\src\module\motion.cpp:151:16: error: 'HOMING_FEEDRATE_XY' was not declared in this scope`

This is for the Ender 3 V1 board.

### Benefits

This allows Marlin to compile for the Ender 3 V1 board.


ellensp:
No. This user used the import branch, not the branch designed for the version of marlin they used.
Import is is for internal use only, users must use a release or bugfix branch.

PR #392, opened 2021-01-06
Update Configuration.h

ils15:
Update some funcionality in MKS Robin Nano file


config/examples/Creality/Ender-3/MKS Robin E3 1.0/Configuration.h
```diff
@@ -70,7 +70,7 @@
 // @section info
 
 // Author info of this build printed to the host during boot and M115
-#define STRING_CONFIG_H_AUTHOR "(thisiskeithb, Ender-3)" // Who made the changes.
+#define STRING_CONFIG_H_AUTHOR "(ils15, Ender-3)" // Who made the changes.
```
thisiskeithb: You can keep my name in the config 🙂

ils15: Ok
but need a double check because motor was inverted in previous release
i believe you fixed that, right?

ils15: you can refer in my marlin for mks robin e3D to check all of this
https://github.com/ils15/MKS-Robin-E3D

PR #407, opened 2021-01-27
Creality Ender 3 Pro SKR 1.4 Turbo (2209) Bltouch

NiiXXiiN:
#### Ender 3 Pro Configurations for SKR 1.4 Turbo with TMC2209 Bugfix CONFIGURATION_H_VERSION 020008

I found no other example that had all these features bellow enabled wich made it a bit hard to make this board work. Also for some reason with the latest stable marlin version i could not make the bltouch sensor work with the probe pins of the board. Only when i switched to bugfix it worked.

This configuration is for an Ender 3 Pro with the following options enabled:
- Motherboard BigTreeTech SKR 1.4 Turbo
- Drivers TMC2209 (sensorless homing enabled)
- BLtouch Bed leveling Sensor (Bilinear Mode 5x5 Grid) (sensor plugged into the probe pins)
- Auto Fillament Load and unload enabled (**_measure your bowden length + extruder up to nozzle and change values!!!_** FILAMENT_CHANGE_UNLOAD_LENGTH and FILAMENT_CHANGE_FAST_LOAD_LENGTH)
- Linear Advance enabled (**_Do a calibration for your correct values!!!_**)
- TFT 35V3 (EXP3 + TFT cables using #define CR10_STOCKDISPLAY or change it to #define REPRAP_DISCOUNT_FULL_GRAPHIC_SMART_CONTROLLER and use EXP1 and EXP2)
- Extruder Auto Fan at 50C (Extruder Fan connected to Pin P2_04 HE1)


HenryWait:
I think I may have ran into the same bug.
prusa clone
SKR 1.4 BTT board
TMC2209 sensorless homing on X,Y
BLtouch 3.1 clone
Marlin 2.0.7.2 

I have tried countless configuration instructions for the bltouch and no matter what the probe will not trigger, I know it is in the correct pins because it attempts to home Z and when unplugged or put in ZMIN it won't attempt home Z. 
I have tried using ZMIN instead of the dedicated probe pins and I get the same behavior, unplugged no home, plugged attempts but trigger does not stop Z from trying to drill through the bed. 

I'll try the bugfix branch because I have tried every combination of configuration on current stable.

[Configuration.zip](https://github.com/MarlinFirmware/Configurations/files/6186130/Configuration.zip)

PR #417, opened 2021-02-17
add YHCB2004 with v4.1 board models  + A30 variants

Vertabreak:
add geeetech YHCB2004 with v4.1 board models + A30 variants


Sh7777777777 (approved):

Vertabreak:
tried to update it to latest as its been sitting here nearly a month might need some additional looking over.

PR #425, opened 2021-02-24
Mks robin e3 new version

ils15:
To alert users
MKS ROBIN E3 has a new version that needs to be compiled with another motherboard. As some were making a mess, I created these two separate files, and a small read explaining the difference between boards


config/examples/Creality/Ender-3/MKS Robin E3/readme.md
```diff
@@ -0,0 +1,12 @@
+NOW MKS has a V1.1 Version of MKS Board 
+
+Need to compile using another board pin
+the new definition for V1.1 is
+#define MOTHERBOARD BOARD_MKS_ROBIN_E3_V1_1 instead   #define MOTHERBOARD BOARD_MKS_ROBIN_E3
```
thisiskeithb: `BOARD_MKS_ROBIN_E3_V1_1` doesn’t exist in the main Marlin repo, so this config won’t work until support is added.

ils15: Latest Bugfix

Screenshot_7

thisiskeithb: Yes, I see it now. I must have been on a different branch when I looked.

These were added in https://github.com/MarlinFirmware/Marlin/pull/20216.

PR #430, opened 2021-03-02
Original Creation of CR-10 V3 Examples

AaronJust:
### Description

Starting with CR-10 V2 as a basis, updated the boot screen, status screen, and configuration to match a CR-10 V3 (i.e. update e-steps and extruder direction to accommodate the direct drive unit).  Also enabled power loss recovery by default.


roydonmalcolm:
where is the link to download and install the latest version of Marlin firmware for my CR-10 V3 ? I can find all sort of write up but not the actual firmware file?

thisiskeithb:
> where is the link to download and install the latest version of Marlin firmware for my CR-10 V3 ? I can find all sort of write up but not the actual firmware file?

https://github.com/MarlinFirmware/Marlin/releases

roydonmalcolm:
Thank you very Much!!1

PR #456, opened 2021-04-05
FLSUN Q5 stock UBL, Jerk+Scurve, LA, extruder and bed PID, delta auto…

Hal9999:
Hi, my first PR :)
I edited the config files for FLSUN Q5 including all the advanced features I would find useful.

### Requirements

* FLSUN Q5 stock
* Calibrate PID for bed and extruder
* Optionally you can execute the "M48 repeatability test to test probe accuracy"
* Then run the probe offset wizard
* Then the Delta auto-calibration routine
* Run the UBL mesh measurement routine
* I calibrated the extruder steps, overriding the stock setting (that to me underextruded also with stock firmware)

### Description

I used the robin_nano_config.txt included in the SD card provided by FLSUN to set most of the parameters in the configuration files, keeping track of where this settings are provided in the original configuration file.
Then I enabled all wizard and settings I evaluated useful.

In terms of speed and acceleration, the configuration is set as stock as possible

### Benefits

FLSUN Q5 configuration for Marlin bugfix

### Related Issues

After G26 ends, the printer doesn't home by herself.

### DIFF
--- a/config/examples/delta/FLSUN/Q5-nano_v1/Configuration.h
+++ b/config/examples/delta/FLSUN/Q5-nano_v1/Configuration.h
@@ -21,8 +21,6 @@
  */
 #pragma once
 
-#define CONFIG_EXAMPLES_DIR "delta/FLSUN/Q5-nano_v1"
-
 /**
  * Configuration.h
  *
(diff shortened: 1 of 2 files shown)


thinkyhead:
Thanks for your contribution! Unfortunately we can't accept PRs directed at auto-generated branches. We make patches to the "import" branches and only later do we regenerate `bugfix-2.0.x` and release branches. Please redo this PR starting with the `import-2.0.x` branch and be careful to target `import-2.0.x` when resubmitting the PR.

PR #463, opened 2021-04-08
correct A20 E0 dir

Vertabreak:
correct A20 E0 dir


Vertabreak:
https://github.com/MarlinFirmware/Marlin/issues/21561#issuecomment-815511635

PR #481, opened 2021-05-03
Added Ultimaker Original configuration example

UMO-printing:
Added configurations for Ultimaker Original

### Requirements

* This is a new (and tested) configuration, so no special requirements are needed

### Description

Examples for a 'New' printer is added. New to configuration examples, that is.  
For some unknown reason a configuration example for the printer that gave birth to the Marlin Firmware, has so far not been included in the configuration examples.

### Benefits

Will make it easier to get started  with Marlin 2.0.x on the Ultimaker Original

### Related Issues

none


thisiskeithb:
You'll need to target the `import-2.0.x` branch or PRs are automatically closed after a force-push to `bugfix-2.0.x`.

UMO-printing:
> You'll need to target the `import-2.0.x` branch or PRs are automatically closed after a force-push to `bugfix-2.0.x`.

Ok, Thank You for the heads up :-) 
I will try to create a pull request towards the `import-2.0.x` branch.

ansonl:
@UMO-printing, do you know if the added Ultimaker Original config files are compatible with the Ultimaker Original Plus or is additional modification needed to get the stock UMO+ heated bed working?

Also are X_BED_SIZE and Y_BED_SIZE values are set to 200 in Configuration.h for safety even though the UMO has a max XY print size of 210x210mm?

UMO-printing:
> @UMO-printing, do you know if the added Ultimaker Original config files are compatible with the Ultimaker Original Plus or is additional modification needed to get the stock UMO+ heated bed working?

I think that you first will need to find out if your UMO+ uses the (green) 1.5.7 PCB that sits on top of an Arduino Mega. 
Of if (more likely) your UMO+ uses the (white) UltiMain 2.1.4 PCB. 

--- 

If your UMO+ uses the (green) 1.5.7  PCB, then I think you need to change a few #defines to get the heated bed working 

I see the: 
  `#define TEMP_SENSOR_BED 0`
the 0 needs to be changed to the type of sensor that is used for the heated bed. 

You might also want to change the: 
 `//#define PIDTEMPBED `
to 
 `#define PIDTEMPBED`
 
 I think that should be it, to activate the heated bed. 

 ---

If your UMO+ does however use the (white) 2.1.4 PCB, then a quick look shows me, that you need to change:

from 
`#define MOTHERBOARD BOARD_ULTIMAKER`
to 
`#define MOTHERBOARD BOARD_ULTIMAIN_2`

And in the file `board.h`, it is suggested that the temperature sensor is number 20

At temperature sensor 20,  I do however read (in `Configuration.h`) :
```
 *    20 : Pt100 with circuit in the Ultimainboard V2.x with mainboard ADC reference voltage = INA826 amplifier-board supply voltage.
 *         NOTES: (1) Must use an ADC input with no pullup. 
                  (2) Some INA826 amplifiers are unreliable at 3.3V so consider using sensor 147, 110, or 21.
```
 
 So it will definitely be important that you first find out which  PCB is on you UMO+

I do however think that we can ignore the Notes, as they are most likely not relevant for the actual genuine 2.1.4 PCB. But those notes are more likely some added info, so it is possible to use the PT100/INA826 combination with other controller boards. 

--- 

> Also are X_BED_SIZE and Y_BED_SIZE values are set to 200 in Configuration.h for safety even though the UMO has a max XY print size of 210x210mm?

No, that is an oversight/error on my part. 

--- 

I am wishing to get access to an UMO+ (with a 2.1.4 PCB), so I can experiment to find the right Marlin settings for a standard configuration example for that printer as well.  But so far I have not found such a printer (locally), that I have been able to test on. 
Perhaps you will succeed with such a task. So an UMO+ example configuration can also be added to the Marlin example configurations?

ansonl:
Thanks for the info, I've created a working firmware for the stock Ultimaker Original Plus using the following configurations for reference:

- [Ultimaker UM Original Plus configuration](https://github.com/Ultimaker/Marlin/tree/Marlin_UM_Original_Plus) 
  - Official Ultimaker Original Plus Marlin v1 configuration
- [Ultimaker UM2 configuration](https://github.com/Ultimaker/Ultimaker2Marlin)
  - Used for heated bed constant values. UMO+ and UM2 heated bed should be similar.
- [gyohng/Marlin2-UltimakerOriginal](https://github.com/gyohng/Marlin2-UltimakerOriginal)
  - Marlin v2 configuration for a user's modified UMO

Your board and thermostat guesses seem to be correct and the UMO+ heated bed appears to be controlled by a MOSFET according to the Ultimainboard v2.x schematics so PIDTEMPBED should be compatible. 

I'm still testing the configuration out and will add it to the repo after a long print.

PR #512, opened 2021-06-05
Add Sunlu S8

InsanityAutomation:
Add Sunlu S8 printer example configurations


thisiskeithb:
@InsanityAutomation: Did your machine really come with a RAMPS board instead of an MKS Base motherboard?

From what I can find online and in reviews, it should be an MKS Base like reported in https://github.com/MarlinFirmware/Configurations/issues/713

InsanityAutomation:
Yup, fan on mine is always on. Just changing the board type there would only add a case light and set mosfet d pin to 7, which the config doesnt use. Functionality described would be using the auto fan pin in config_adv.

thisiskeithb:
Thanks!

I'll create a new config since these also shipped with MKS Base boards as well.

InsanityAutomation:
Mine is an oddball, not an off the shelf board. may have been based on the mks base board, but also the board change alone does nothing for the config so irrelevant. If there is something defining an auto fan pin, that could be added to the same config and if its wired always on like mine its safely doing nothing.

thisiskeithb:
>  If there is something defining an auto fan pin, that could be added to the same config

Yep. I'm on it.

> and if its wired always on like mine its safely doing nothing.

It's not. Please see https://github.com/MarlinFirmware/Configurations/issues/713

PR #518, opened 2021-06-09
fix various BLTOUCH conflicts in example configs  

ellensp:
### Description

Part of PR https://github.com/MarlinFirmware/Marlin/pull/22086

#define BLTOUCH and #define DEACTIVATE_SERVOS_AFTER_MOVE, should error
#define BLTOUCH and #define Z_MIN_PROBE_ENDSTOP_INVERTING true, should error
if ENABLED(Z_MIN_PROBE_USES_Z_MIN_ENDSTOP_PIN)
  #define BLTOUCH and #define Z_MIN_ENDSTOP_INVERTING true, should error

Fix example configs that break these rules 

### Benefits

Corrected Configuration.h

### Related Issues
https://github.com/MarlinFirmware/Marlin/pull/22086

PR #520, opened 2021-06-13
fix Z_MIN_PROBE_ENDSTOP_INVERTING set true when bltouch is enabled

ellensp:
### Requirements

Enable CR10V2_BLTOUCH in Creality/CR-10 V2/Configuration.h
Missed this one, due to config using custom CR10V2_BLTOUCH define. 

### Description

This incorrectly sets Z_MIN_PROBE_ENDSTOP_INVERTING true
Updated to be false

### Benefits

no #error "BLTOUCH requires Z_MIN_PROBE_ENDSTOP_INVERTING set to false. Please update your Configuration.h file."

### Related Issues
https://github.com/MarlinFirmware/Configurations/pull/518

### DIFF
--- a/config/examples/Creality/CR-10 V2/Configuration.h
+++ b/config/examples/Creality/CR-10 V2/Configuration.h
@@ -818,7 +818,7 @@
 #define I_MAX_ENDSTOP_INVERTING false // Set to true to invert the logic of the endstop.
 #define J_MAX_ENDSTOP_INVERTING false // Set to true to invert the logic of the endstop.
 #define K_MAX_ENDSTOP_INVERTING false // Set to true to invert the logic of the endstop.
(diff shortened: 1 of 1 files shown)


rhapsodyv:
Maybe worth putting a comment in the configuration.h, on top of Z_MIN_PROBE_ENDSTOP_INVERTING, saying it's requeriment for bltouch too...

thinkyhead:
Not just this as a correction…?
```cpp
#define Z_MIN_PROBE_ENDSTOP_INVERTING DISABLED(CR10V2_BLTOUCH) // Set to true to invert the logic of the probe.
```

ellensp:
With DISABLED Z_MIN_PROBE_ENDSTOP_INVERTING is true when CR10V2_BLTOUCH is disabled. it should be false then too.

CRCinAU:
@thinkyhead Any chance this can be put into the bugfix-2.0.x branch too?

Just so that the example config works again.

PR #523, opened 2021-06-14
Creality CR10 V2: BLTouch disabled build fix

sitic:
### Requirements

The BLTouch is an optional upgrade for the  Creality CR10 v2.

Currently the build fails when CR10V2_BLTOUCH is disabled:

"G29_RETRY_AND_RECOVER requires AUTO_BED_LEVELING_3POINT, LINEAR, or BILINEAR."

### Description

Only enable G29_RETRY_AND_RECOVER when CR10V2_BLTOUCH is enabled.


thisiskeithb:
There’s a related discussion in https://github.com/MarlinFirmware/Configurations/pull/520

PR #530, opened 2021-06-26
Update Touch Screen calibration parameters

ils15:
Hi
Some users passes new touch calibration
please update config a file
thanks


rhapsodyv:
If calibration values are changing from users to users, it indicates that they have minor hardware differences that affect calibration (it’s very sensible).

So, instead of changing the values, consider that the best you can do is remove these default values and turn on auto save.

Then users will need to calibrate just once when boot and it will be saved in theirs eeprom.

rhapsodyv:
I had similar issues with x5sa config, and had to remove defaults.

ils15:
> I had similar issues with x5sa config, and had to remove defaults.

Robin ts35 has a problem in display when it starts Calibration process
I need to investigate if new marlin solved it, but in 2.0.8 I made a compilation without calibration and all users cannot calibrate printer using display (M995 only works when they plug and send gcode using pronterface)

rhapsodyv:
Calibration screen only shows up in the boot process if all calibration values are 0. Maybe there’s some dirty somewhere

PR #533, opened 2021-06-30
Fix default configs with 'BLTOUCH' and 'Z_MIN_ENDSTOP_INVERTING' => true

slowbro:
### Description

`BLTOUCH` requires `Z_MIN_ENDSTOP_INVERTING false` but some example configs have it set to 'true'

### Benefits

Fixes improper example configs.

### Related Issues

Mentioned in MarlinFirmware/Marlin#22272 and possibly elsewhere.


DavBfr:
Please update this one too:

https://github.com/MarlinFirmware/Configurations/blob/import-2.0.x/config/examples/Creality/CR-10%20V3/Configuration.h#L823

slowbro:
Thanks, added that, and some others that use the same pattern.

CRCinAU:
Thanks for lodging this - I just got hit by this yet again hahaha

PR #539, opened 2021-07-10
Fix Garbled Tevo Tornado Bootscreen

thisiskeithb:
### Description

Set correct bootscreen width to prevent a garbled bootscreen on the Tevo Tornado. I also removed the custom timeout since it was too short.

### Benefits

No more garbled bootscreen!

image

### Related Issues

https://github.com/MarlinFirmware/Marlin/issues/22326
https://github.com/MarlinFirmware/Marlin/issues/21382


yet-another-average-joe:
Are you sure that removing CUSTOM_BOOTSCREEN_TIMEOUT is the right thing to do ? Keeping it in _Bootscreen.h could be better. Or just comment it.

It's been there since the very first version of the Tornado firmware, and was set to 3 seconds. The new bitmap saves 64 bytes, but removing CUSTOM_BOOTSCREEN_TIMEOUT gives no benefit, and hides a customization very deep in marlinui_DOGM.cpp

[Tevo Tornado v1.0 _Bootscreen.h.zip](https://github.com/MarlinFirmware/Configurations/files/6795161/Tevo.Tornado.v1.0._Bootscreen.h.zip)

thisiskeithb:
> Are you sure that removing CUSTOM_BOOTSCREEN_TIMEOUT is the right thing to do ?

Yep.

thinkyhead:
> Are you sure that removing `CUSTOM_BOOTSCREEN_TIMEOUT` is the right thing to do ?

It should be left in if the shorter value improves the bootscreen experience. It will fall back to 2.5 seconds without it.

PR #541, opened 2021-07-14
Add TDK NTCG104LH104KT1 NTC100K Thermistor

thisiskeithb:
### Description

Motherboard thermistor used on the Ultimachine Einsy Rambo.

### Benefits

Einsy Rambo users will be able to monitor motherboard temps (once we figure out why temps aren't displaying correctly).

### Related Issues

Sister PR to https://github.com/MarlinFirmware/Marlin/pull/22350


slowbro:
Curious, why such a high number? Other things in the thousands seem to be 'variants' of other thermistors. I assume to make it stand apart since it would be generally only used on the mobo

thisiskeithb:
> Curious, why such a high number? Other things in the thousands seem to be 'variants' of other thermistors. I assume to make it stand apart since it would be generally only used on the mobo

It's what Prusa used in their firmware, so I went with that.

PR #550, opened 2021-08-04
Updated working config for Creality CR-10 V3

PyreDev:
Hello,

I would like to share my working config to run Marlin 2.0 on a Creality CR-10 V3 printer if it can helps others. I suggest to all CR-10 v3 owners to upgrade for Marlin 2.0 as it is adding a couple of very useful features for this machine that comes from Creality with a v1.1.6 old Marlin.
I feel like I have a new machine that's great !

Global changes : 

- Corrected the max nozzle temp as the original Creality FW. In the existing config it's limited to 240°C (260°C in Original FW)
- Residency temp time was too low and was making the machine doing a "false start" sometimes.
- The max speeds have been set down to suitable values for this machine stock.
- The max acceleration setting set down to the sames values than Original CR-10 V3 FW.
- The defauts homing/probing/park speeds have been speed up a little, this have been tested to have faster smooth movement when homing, leveling etc.. with a stock machine, accuracy of the BLTOUCH has been tested as well, with a comfortable safety margin. 
- Correction of the Change filament fonction that was set for a Bowden extruder.
- Adding and configure Level Corner Function
- Adding and configure Tramming Wizard fonction
- Adding custom menu with PARK nozzle command.
- Printing area was not centered
- Probing area was not centered to Printing area.
- Sd menu was lagging a lot.
- There was an unusable FAN2 in the menus
- CR-10 V3 status logo updated with standard LCD characters
- Added sound menu to be able to chose buzze on or off


config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -76,7 +76,7 @@
 // @section info
 
 // Author info of this build printed to the host during boot and M115
-#define STRING_CONFIG_H_AUTHOR "Aaron Just, CR10-V3)" // Who made the changes.
+#define STRING_CONFIG_H_AUTHOR "PYRE" // Who made the changes.
```
PyreDev: Changed to me !

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -518,11 +518,11 @@
 //#define MAX31865_SENSOR_OHMS_1      100
 //#define MAX31865_CALIBRATION_OHMS_1 430
 
-#define TEMP_RESIDENCY_TIME          2  // (seconds) Time to wait for hotend to "settle" in M109
+#define TEMP_RESIDENCY_TIME          5  // (seconds) Time to wait for hotend to "settle" in M109
```
PyreDev, line 521: 2 sec was too low. Sometimes after starting moving, the machine was stopping again to re-heat

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -518,11 +518,11 @@
...
 #define TEMP_WINDOW                  1  // (°C) Temperature proximity for the "temperature reached" timer
 #define TEMP_HYSTERESIS              3  // (°C) Temperature proximity considered "close enough" to the target
 
-#define TEMP_BED_RESIDENCY_TIME      5  // (seconds) Time to wait for bed to "settle" in M190
+#define TEMP_BED_RESIDENCY_TIME      8  // (seconds) Time to wait for bed to "settle" in M190
```
PyreDev, line 525: As the residency time was too short for the hot end I also increase that one

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -562,7 +562,7 @@
 // Above this temperature the heater will be switched off.
 // This can protect components from overheating, but NOT from shorts and failures.
 // (Use MINTEMP for thermistor short/failure protection.)
-#define HEATER_0_MAXTEMP 255
+#define HEATER_0_MAXTEMP 275
```
PyreDev, line 565: To be able to heat at 260 according original FW

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -926,18 +926,18 @@
  *                                      X, Y, Z [, I [, J [, K]]], E0 [, E1[, E2...]]
  * Updated E Steps to 382.14 for CR10-V3 Direct Drive
  */
-#define DEFAULT_AXIS_STEPS_PER_UNIT   { 80.00, 80.00, 400.00, 382.14 }
+#define DEFAULT_AXIS_STEPS_PER_UNIT   { 80.00, 80.00, 400.00, 407 }
```
PyreDev, line 929: 407 works fine and is also close to the E3D extruder theoric spec.

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -926,18 +926,18 @@
...
 
 /**
  * Default Max Feed Rate (mm/s)
  * Override with M203
  *                                      X, Y, Z [, I [, J [, K]]], E0 [, E1[, E2...]]
  */
-#define DEFAULT_MAX_FEEDRATE          { 750, 750, 15, 75 }
+#define DEFAULT_MAX_FEEDRATE          { 200, 200, 14, 25 }
```
PyreDev, line 936: X & Y : max value for smooth operation with stock CR-10 V3
Extruder : as original Creality FW

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -926,18 +926,18 @@
...
 
-//#define LIMITED_MAX_FR_EDITING        // Limit edit via M203 or LCD to DEFAULT_MAX_FEEDRATE * 2
+#define LIMITED_MAX_FR_EDITING        // Limit edit via M203 or LCD to DEFAULT_MAX_FEEDRATE * 2
 #if ENABLED(LIMITED_MAX_FR_EDITING)
-  #define MAX_FEEDRATE_EDIT_VALUES    { 600, 600, 10, 50 } // ...or, set your own edit limits
+  #define MAX_FEEDRATE_EDIT_VALUES    { 200, 200, 14, 25 } // ...or, set your own edit limits
```
PyreDev, line 940: X & Y : max value for smooth operation with stock CR-10 V3
Extruder : as original Creality FW

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -946,7 +946,7 @@
  * Override with M201
  *                                      X, Y, Z [, I [, J [, K]]], E0 [, E1[, E2...]]
  */
-#define DEFAULT_MAX_ACCELERATION      { 2000, 2000, 100, 5000 }
+#define DEFAULT_MAX_ACCELERATION      { 500, 500, 100, 1000 }
```
PyreDev, line 949: Same value as original FW settings

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -961,9 +961,9 @@
  *   M204 R    Retract Acceleration
  *   M204 T    Travel Acceleration
  */
-#define DEFAULT_ACCELERATION           800    // X, Y, Z and E acceleration for printing moves
-#define DEFAULT_RETRACT_ACCELERATION  1000    // E acceleration for retracts
-#define DEFAULT_TRAVEL_ACCELERATION    800    // X, Y, Z acceleration for travel (non printing) moves
+#define DEFAULT_ACCELERATION          500    // X, Y, Z and E acceleration for printing moves
+#define DEFAULT_RETRACT_ACCELERATION  500    // E acceleration for retracts
+#define DEFAULT_TRAVEL_ACCELERATION  1000    // X, Y, Z acceleration for travel (non printing) moves
```
PyreDev, line 966: Sames values than original FW

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -1187,17 +1187,17 @@
  *     |    [-]    |
  *     O-- FRONT --+
  */
-#define NOZZLE_TO_PROBE_OFFSET { 47, 0, 0 }
+#define NOZZLE_TO_PROBE_OFFSET { 46, 2, 0 }
 
 // Most probes should stay away from the edges of the bed, but
 // with NOZZLE_AS_PROBE this can be negative for a wider probing area.
-#define PROBING_MARGIN 10
+#define PROBING_MARGIN 45
```
PyreDev: Adjust for the more extended probing area possible with cr-10 v3

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -1187,17 +1187,17 @@
...
 
 // X and Y axis travel speed (mm/min) between probes
-#define XY_PROBE_FEEDRATE (150*60)
+#define XY_PROBE_FEEDRATE (200*60)
```
PyreDev, line 1198: To speed up Levelings

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -1187,17 +1187,17 @@
...
 
 // Feedrate (mm/min) for the first approach when double-probing (MULTIPLE_PROBING == 2)
-#define Z_PROBE_FEEDRATE_FAST (6*60)
+#define Z_PROBE_FEEDRATE_FAST (14*60)
```
PyreDev, line 1201: Max values for smooth operation and good BLtouch accuracy

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -1256,11 +1256,11 @@
  *     But: `M851 Z+1` with a CLEARANCE of 2  =>  2mm from bed to nozzle.
  */
 #define Z_CLEARANCE_DEPLOY_PROBE   10 // Z Clearance for Deploy/Stow
-#define Z_CLEARANCE_BETWEEN_PROBES  5 // Z Clearance between probe points
-#define Z_CLEARANCE_MULTI_PROBE     5 // Z Clearance between multiple probes
-#define Z_AFTER_PROBING           5 // Z position after probing is done
+#define Z_CLEARANCE_BETWEEN_PROBES  3 // Stock: 5 - Z Clearance between probe points
+#define Z_CLEARANCE_MULTI_PROBE     3 // Stock: 5 - Z Clearance between multiple probes
```
PyreDev, line 1261: This save a little bit of time for levelings. Working with standard Creality BLTouch mount

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -1380,12 +1380,12 @@
 // @section machine
 
 // The size of the printable area
-#define X_BED_SIZE 310 // Nozzle is at X0 when homed
-#define Y_BED_SIZE 310 // Nozzle is at Y0 when homed
+#define X_BED_SIZE 300 // Nozzle is at X0 when homed
+#define Y_BED_SIZE 300 // Nozzle is at Y0 when homed
```
PyreDev, line 1385: Correct print area according to spec

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -1380,12 +1380,12 @@
...
 
 // Travel limits (mm) after homing, corresponding to endstop positions.
-#define X_MIN_POS 0
-#define Y_MIN_POS 0
+#define X_MIN_POS -5 // -5 instead of 0 to center the print area (CR-10 v3)
+#define Y_MIN_POS -5 // -5 instead of 0 to center the print area (CR-10 v3)
```
PyreDev, line 1389: To center the print area

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -1567,7 +1567,7 @@
 /**
  * Auto-leveling needs preheating
  */
-//#define PREHEAT_BEFORE_LEVELING
+#define PREHEAT_BEFORE_LEVELING
```
PyreDev: For improved accuracy, but it can be confusing as you don't get any clear message that the printer is waiting for heating

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -1694,14 +1694,14 @@
 #endif
 
 // Add a menu item to move between bed corners for manual bed adjustment
-//#define LEVEL_BED_CORNERS
+#define LEVEL_BED_CORNERS
```
PyreDev, line 1698: Very useful option

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -1694,14 +1694,14 @@
...
 
 #if ENABLED(LEVEL_BED_CORNERS)
-  #define LEVEL_CORNERS_INSET_LFRB { 30, 30, 30, 30 } // (mm) Left, Front, Right, Back insets
+  #define LEVEL_CORNERS_INSET_LFRB { 45, 45, 45, 45 } // (mm) Left, Front, Right, Back insets
```
PyreDev, line 1701: Margin adapted to cr-10 v3 bed

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -1737,7 +1737,7 @@
 // @section homing
 
 // The center of the bed is at (X=0, Y=0)
-//#define BED_CENTER_AT_0_0
+#define BED_CENTER_AT_150_150
```
PyreDev: center of the 300x300 print surface

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -1765,7 +1765,7 @@
 #endif
 
 // Homing speeds (mm/min)
-#define HOMING_FEEDRATE_MM_M { (50*60), (50*60), (6*60) }
+#define HOMING_FEEDRATE_MM_M { (80*60), (80*60), (10*60) }
```
PyreDev, line 1769: Speed increased a little but not up to the max

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -1902,12 +1902,12 @@
 
 #if ENABLED(NOZZLE_PARK_FEATURE)
   // Specify a park position as { X, Y, Z_raise }
-  #define NOZZLE_PARK_POINT { (X_MIN_POS + 5), (Y_MAX_POS - 5), 10 }
+  #define NOZZLE_PARK_POINT { (X_MIN_POS + 5), (Y_MAX_POS - 5), 100 }
```
PyreDev, line 1906: I like to park with the nozzle accessible for inspection/cleaning

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -1902,12 +1902,12 @@
...
   //#define NOZZLE_PARK_X_ONLY          // X move only is required to park
   //#define NOZZLE_PARK_Y_ONLY          // Y move only is required to park
   #define NOZZLE_PARK_Z_RAISE_MIN   2   // (mm) Always raise Z by at least this distance
-  #define NOZZLE_PARK_XY_FEEDRATE 80   // (mm/s) X and Y axes feedrate (also used for delta Z axis)
-  #define NOZZLE_PARK_Z_FEEDRATE    5   // (mm/s) Z axis feedrate (not used for delta printers)
+  #define NOZZLE_PARK_XY_FEEDRATE 200   // (mm/s) X and Y axes feedrate (also used for delta Z axis)
+  #define NOZZLE_PARK_Z_FEEDRATE   14   // (mm/s) Z axis feedrate (not used for delta printers)
```
PyreDev, line 1911: We can go fast for this

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -2833,7 +2833,7 @@
 
 // Set number of user-controlled fans. Disable to use all board-defined fans.
 // :[1,2,3,4,5,6,7,8]
-//#define NUM_M106_FANS 1
+#define NUM_M106_FANS 1
```
PyreDev, line 2837: Only one controllable fan on this machine.
If not defined then we have 2 fan in the menus

config/examples/Creality/CR-10 V3/Configuration_adv.h
```diff
@@ -928,11 +928,11 @@
 //
 // Add the G35 command to read bed corners to help adjust screws. Requires a bed probe.
 //
-//#define ASSISTED_TRAMMING
+#define ASSISTED_TRAMMING
```
PyreDev, line 931: Useful option

config/examples/Creality/CR-10 V3/Configuration_adv.h
```diff
@@ -928,11 +928,11 @@
...
 #if ENABLED(ASSISTED_TRAMMING)
 
   // Define positions for probe points.
-  #define TRAMMING_POINT_XY { {  20, 20 }, { 180,  20 }, { 180, 180 }, { 20, 180 } }
+  #define TRAMMING_POINT_XY { {  45, 45 }, { 255,  45 }, { 255, 255 }, { 45, 255 } }
```
PyreDev: Adapted for maximum probing area on the CR-10 V3 (and going as close as possible to the bed screws)

config/examples/Creality/CR-10 V3/Configuration_adv.h
```diff
@@ -943,7 +943,7 @@
   #define RESTORE_LEVELING_AFTER_G35    // Enable to restore leveling setup after operation
   //#define REPORT_TRAMMING_MM          // Report Z deviation (mm) for each point relative to the first
 
-  //#define ASSISTED_TRAMMING_WIZARD    // Add a Tramming Wizard to the LCD menu
+  #define ASSISTED_TRAMMING_WIZARD    // Add a Tramming Wizard to the LCD menu
```
PyreDev, line 946: Useful option

config/examples/Creality/CR-10 V3/Configuration_adv.h
```diff
@@ -1247,7 +1247,7 @@
   //#define TURBO_BACK_MENU_ITEM
 
   // Add a mute option to the LCD menu
-  //#define SOUND_MENU_ITEM
+  #define SOUND_MENU_ITEM
```
PyreDev, line 1250: VERY nice improvement !! Was not available with the original creality FW

config/examples/Creality/CR-10 V3/Configuration_adv.h
```diff
@@ -1382,14 +1382,14 @@
    */
   #define POWER_LOSS_RECOVERY
   #if ENABLED(POWER_LOSS_RECOVERY)
-    #define PLR_ENABLED_DEFAULT   false // Power Loss Recovery enabled by default. (Set with 'M413 Sn' & M500)
+    #define PLR_ENABLED_DEFAULT    true // Power Loss Recovery enabled by default. (Set with 'M413 Sn' & M500)
```
PyreDev, line 1385: PLR enabled by defaut

config/examples/Creality/CR-10 V3/Configuration_adv.h
```diff
@@ -1435,8 +1435,8 @@
     #define SDSORT_GCODE       false  // Allow turning sorting on/off with LCD and M34 G-code.
     #define SDSORT_USES_RAM    true   // Pre-allocate a static array for faster pre-sorting.
     #define SDSORT_USES_STACK  true   // Prefer the stack for pre-sorting to give back some SRAM. (Negated by next 2 options.)
-    #define SDSORT_CACHE_NAMES false  // Keep sorted items in RAM longer for speedy performance. Most expensive option.
-    #define SDSORT_DYNAMIC_RAM false  // Use dynamic allocation (within SD menus). Least expensive option. Set SDSORT_LIMIT before use!
+    #define SDSORT_CACHE_NAMES true   // Keep sorted items in RAM longer for speedy performance. Most expensive option.
```
PyreDev, line 1438: Working very nice for me. It has definitionally improved the navigation reactivity in the sdcard browser. Before it was really anoying with a lot of lag.

config/examples/Creality/CR-10 V3/Configuration_adv.h
```diff
@@ -1459,7 +1459,7 @@
    * This feature must be enabled with "M540 S1" or from the LCD menu.
    * To have any effect, endstops must be enabled during SD printing.
    */
-  //#define SD_ABORT_ON_ENDSTOP_HIT
+  #define SD_ABORT_ON_ENDSTOP_HIT
```
PyreDev: If a endstop is hit during print that's not normal and me be safer to stop then, to avoid mechanical damage of the machine

config/examples/Creality/CR-10 V3/Configuration_adv.h
```diff
@@ -2391,7 +2391,7 @@
                                                   // This short retract is done immediately, before parking the nozzle.
   #define FILAMENT_CHANGE_UNLOAD_FEEDRATE     40  // (mm/s) Unload filament feedrate. This can be pretty fast.
   #define FILAMENT_CHANGE_UNLOAD_ACCEL        25  // (mm/s^2) Lower acceleration may allow a faster feedrate.
-  #define FILAMENT_CHANGE_UNLOAD_LENGTH      435  // (mm) The length of filament for a complete unload.
+  #define FILAMENT_CHANGE_UNLOAD_LENGTH       40  // (mm) The length of filament for a complete unload.
```
PyreDev, line 2394: It's a direct extruder on this machine

config/examples/Creality/CR-10 V3/Configuration_adv.h
```diff
@@ -2400,7 +2400,7 @@
                                                   // 0 to disable start loading and skip to fast load only
   #define FILAMENT_CHANGE_FAST_LOAD_FEEDRATE  30  // (mm/s) Load filament feedrate. This can be pretty fast.
   #define FILAMENT_CHANGE_FAST_LOAD_ACCEL     15  // (mm/s^2) Lower acceleration may allow a faster feedrate.
-  #define FILAMENT_CHANGE_FAST_LOAD_LENGTH   430  // (mm) Load length of filament, from extruder gear to nozzle.
+  #define FILAMENT_CHANGE_FAST_LOAD_LENGTH     0  // (mm) Load length of filament, from extruder gear to nozzle.
```
PyreDev, line 2403: direct extruder

config/examples/Creality/CR-10 V3/Configuration_adv.h
```diff
@@ -3722,20 +3722,20 @@
  */
 
 // Custom Menu: Main Menu
-//#define CUSTOM_MENU_MAIN
+#define CUSTOM_MENU_MAIN
```
PyreDev, line 3725: Just to add a quick command for parking

config/examples/Creality/CR-10 V3/Configuration_adv.h
```diff
@@ -4155,10 +4155,10 @@
...
   #define SERVICE_WARNING_BUZZES  3
   // Activate up to 3 service interval watchdogs
-  //#define SERVICE_NAME_1      "Service S"
-  //#define SERVICE_INTERVAL_1  100 // print hours
-  //#define SERVICE_NAME_2      "Service L"
-  //#define SERVICE_INTERVAL_2  200 // print hours
+  #define SERVICE_NAME_1      "Service 50h"
+  #define SERVICE_INTERVAL_1  50 // print hours
+  #define SERVICE_NAME_2      "Service 300h"
+  #define SERVICE_INTERVAL_2  300 // print hours
```
PyreDev, line 4161: Defined light & deep service intervals

config/examples/Creality/CR-10 V3/_Statusscreen.h
```diff
@@ -1,68 +1,69 @@
...
+  B00000000,B00000000,B00000000,B00000000,
+  B00000000,B00000000,B00000000,B00000000,
+  B00000000,B01000101,B11110000,B00000000,
+  B00000000,B01000100,B00100000,B00000000,
+  B00000000,B01000100,B01000000,B00000000,
+  B00000000,B01000100,B00100000,B00000000,
+  B00000000,B01000100,B00010000,B00000000,
+  B00000000,B00101001,B00010000,B00000000,
+  B00000000,B00010000,B11100000,B00000000,
+};
```
PyreDev: Standard LCD chars for the Status Screen

PyreDev (commented):
Changes commented

config/examples/Creality/CR-10 V3/Configuration.h
```diff
@@ -1452,8 +1452,8 @@
   #define FIL_RUNOUT_PIN 2 // Creality CR10-S stock sensor
 
   #define FIL_RUNOUT_STATE     LOW        // Pin state indicating that filament is NOT present.
-  #define FIL_RUNOUT_PULLUP               // Use internal pullup for filament runout pins.
-  //#define FIL_RUNOUT_PULLDOWN           // Use internal pulldown for filament runout pins.
+  //#define FIL_RUNOUT_PULLUP               // Use internal pullup for filament runout pins.
+  #define FIL_RUNOUT_PULLDOWN           // Use internal pulldown for filament runout pins.
```
PyreDev, line 1457: For me it makes more sense to have a pulldown here as the RUNOUT state is LOW.
But neither with pullup or pulldown i'm not able to get a runout condition by unplugging the sensor during a print.
If you try with real filament in the sensor, in and out, it works fine.

config/examples/Creality/CR-10 V3/Configuration.h
thisiskeithb: You shouldn't remove config author credits.

PyreDev: Hi, sorry I was assuming I could with the comment. I would not have changed it with "Who create the config file" as comment.
I just changed it back to original in my repository.
Should I do an extra pull request for that or is there a way to modify this within this pull request ? I'm quite new on Github I'm still discovering

PyreDev: I've my answer I think, it seems to be automatically applied to the pull request

config/examples/Creality/CR-10 V3/Configuration.h
Vertabreak: this change would shrink the probing area by 90mm not extend it, probing margin moves the probing area in from all edges of the bed this many mm. this feature is largely over used when the build volume should have been directly changed allowing for the proper area to be probed. generally the only probe that would require the use of margin is any probe that does not physically touch the bed as they do not read near the edge.

config/examples/Creality/CR-10 V3/Configuration.h
Vertabreak: in some cases additional speed will reduce accuracy beyond the acceptable 50 micron deviation.

config/examples/Creality/CR-10 V3/Configuration.h
Vertabreak: this option is only for Delta and SCARA printers that require x0 y0 to be in the center.
changing this define the way you have literally does nothing as it links no where.

config/examples/Creality/CR-10 V3/Configuration_adv.h
Vertabreak: enabling this option could result in false positive readings due to vibrations or for example a pet touching the end stop that will result in failed prints.

config/examples/Creality/CR-10 V3/Configuration.h
PyreDev: I have the machine next to me I can confirm my probing  area is not shrinked by 90mm.
This because of nozzle and probe offset. 
If i'm hitting the endstop X, the probe is already +/- 40mm in the print area

config/examples/Creality/CR-10 V3/Configuration.h
PyreDev: It has been test with multiple attempts at same speeds, multiple speeds, and always doing M48 probe test.
With those speeds, the max deviation from 5 attempts was about 0,002498

config/examples/Creality/CR-10 V3/Configuration.h
Vertabreak: probe margin only effects mesh bed leveling probing area, i do not believe you fully understand a number of the changes made.

config/examples/Creality/CR-10 V3/Configuration.h
PyreDev: Ok I will check this, maybe we should add a comment in that way then

config/examples/Creality/CR-10 V3/Configuration.h
Vertabreak: just because something works for you does not mean it will be the same for someone else there are many variants of the bltouch and many other types of probes.

config/examples/Creality/CR-10 V3/Configuration.h
PyreDev: If the probe is already 40mm IN the print area when the X carriage is pushing the endstop, how could you probe less than 40mm ?

config/examples/Creality/CR-10 V3/Configuration.h
PyreDev: In was test I was going up to 22*60 with still 0,0036 max dev on 5 attemps. So I think it's a fair margin.

config/examples/Creality/CR-10 V3/Configuration.h
Vertabreak: i don't think you are understanding the relation of probe offsets and probe_margin and you are referencing things that are unrelated. any probe that touches the bed has no need for probe_margin at all and should rely only on build volume defines.

config/examples/Creality/CR-10 V3/Configuration_adv.h
PyreDev: Hopefully I have no pet then !
But If my pet is able to go within the printer when it's running I still think it's safer to stop. 

And you still need to activate it in the LCD menu. on/off as you want.

config/examples/Creality/CR-10 V3/Configuration.h
Vertabreak: there are comments in place you just did not back trace the define to see them.

config/examples/Creality/CR-10 V3/Configuration.h
PyreDev: But If I put 40, then I'm hitting the endstop during the leveling. How do you explain that ?

Vertabreak:
in general example configs are intended to be a starting point with generalized settings that will work for most rather then extreme specialization that will only work for a few.

config/examples/Creality/CR-10 V3/Configuration_adv.h
Vertabreak: a number of these settings changes are of a personal preference nature and should remain on your fork.

PyreDev:
> 
> 
> in general example configs are intended to be a starting point with generalized settings that will work for most rather then extreme specialization that will only work for a few.

In the existing exemple : 
- You can't heat up to machine spec 240max (normally 260), short for PETG
- Residency temp time was too low and was making the machine doing a "false start"
- If you go to the limit speed/acceleration set you will broke something.
- New nice feature were not configured (filament change enabled and set to a bowden, that's surprising the first time you test with direct drive...), Level corners, level wizard
- Printing area was not centered
- Probing area was not centered to Printing area.
- Sd menu was lagging a lot.
- There was an unusable FAN2 in the menus


So the intention is really not to post my "tuned" config, but I really thing that someone who has the CR-10 V3 will enjoy Marlin 2.0 with this config as it is, more than the existing. Then up to anybody to tune his own ?

Vertabreak:
standard ptfe melts and off gas's at 250C lowering it was likely for safety.
residency temp settings will be varied from heater to heater to the environment and ambient temp.
speeds and accelerations are often totally overridden by the slicer making marlins settings totally moot. 
the filament change feature is generally left unconfigured in most configs and these features are optional.
due to variations of builds from the manufacturer there can never be a perfect settings for all.
many people totally disable sdprinting and perhaps you had alot of files on the card when generally only the file being printed is the only file on the card. 

the changes presented are largely personal flavor changes. 
due to variations from one machine to another even the same model it is nearly impossible to have a perfect setting for anything.
it is common for company's to swap out parts in every new batch maybe they found a part that was a few cents cheaper or the assembly team miss drilled the holes by a few mm any number of things lead to variation.
changing config files is = to little more then pushing some buttons or flipping some switch's in marlin that anyone can do. 

head developer thinkyhead will make his ruling on the matter as he holds final judgement in all things.

PyreDev:
> 
> 
> standard ptfe melts and off gas's at 250C lowering it was likely for safety.
> residency temp settings will be varied from heater to heater to the environment and ambient temp.
> speeds and accelerations are often totally overridden by the slicer making marlins settings totally moot.
> the filament change feature is generally left unconfigured in most configs and these features are optional.
> due to variations of builds from the manufacturer there can never be a perfect settings for all.
> many people totally disable sdprinting and perhaps you had alot of files on the card when generally only the file being printed is the only file on the card.
> 
> the changes presented are largely personal flavor changes.
> due to variations from one machine to another even the same model it is nearly impossible to have a perfect setting for anything.
> it is common for company's to swap out parts in every new batch maybe they found a part that was a few cents cheaper or the assembly team miss drilled the holes by a few mm any number of things lead to variation.
> changing config files is = to little more then pushing some buttons or flipping some switch's in marlin that anyone can do.



250 would have been fine for me, I was initially wanting to reach 245, the recommended temp from my filament manufacturer.
Be able to print PETG was one of the reason I choose this machine.

The filament change was already enabled, with the bowden settings.
So I strongly suggest you to disable this in the existing configuration that should fit for everyone...
The first time it causes me to dismount my extruder to clean up the mess after it tries to load 435mm...

There isn't a lot of configuration / documentation existing for this machine actually, So i was thinking this could help the community, as the example config is not suitable on a couple of aspects. Especially for those who don't like/want/know to put their nose into configuration files and coding.
But it seems not :)

Vertabreak:
some of the proposed changes i have no issues with.
that being said others suggest to me that there is a lack of understanding present.
I'm simply adding to the peer review process adding my thoughts on things that are a bit suspect.
the intention for us all is to help the community but good intentions don't always lead to good results.
say every change you suggested where accepted just looking now i can see more then a dozen use cases that will be negativity impacted.

PyreDev:
I can understand. I tried to put just the options that I would have like to see when I bought the Machine based on their added value for the user with that specific machine (as it's a machine specific config).

Exemple having the custom menu with the park command, normally you don't have access to this without pronterface and directly enter the command, when it something so much useful !
Also if you have one already Custom command defined, may be the user will say "Oh a custom menu ? So I can customize things ? I should have a look into that !" leading to discover all the possibilities of Marlin.

Let me know if would like any video about the probing area margin. when it's set with 45mm margin, it's probing at 45mm and all is working fine.

The endstops it's because I've some industrial electricity/mechanical background and disabling the endstops of a machine it's something just terrible, it's something totally forbidden. There aren't no normal reason to have an endstop triggered when running. No matter if it's because of a broken sensor, a debris, a cat, a failed print, an axis de-synchronization, a kid, in all cases it means that something is potentially wrong.

config/examples/Creality/CR-10 V3/Configuration.h
PyreDev: I have looked for a comment like that but have not found. 
Can you indicate the line please ?

Vertabreak: Marlin/src/inc/Conditionals_post.h
Line 125

config/examples/Creality/CR-10 V3/Configuration.h
Vertabreak: its a combinations of other suspect settings that result in a behavior you consider abnormal. 
i have 4 printers with a 40mm probe offsets and have probe_margin 100% disabled not even set to 0.
i am hold a firm stance that both margin and inset should never be used with a probe that physically touches the bed a proper well define build volume is the solution.
the over use of both margin and inset for less then intended reasons has lead to many other issues that in turn get reported a bugs that are really user error.

config/examples/Creality/CR-10 V3/Configuration.h
PyreDev: I will change it back to original and will just add the comment from Conditionals_post.h by the way.

The comment is : 
// Require 0,0 bed center for Delta and SCARA

It don't explicitly tells that it's not needed at all in other cases. But with what I understand now, it's not needed to have it defined in Configuration.h

If it is already automatically defined in Conditionals_post.h , can we not just remove those 2 lines from Configuration.h ?

config/examples/Creality/CR-10 V3/Configuration.h
PyreDev: Actually both will work. But with 10mm margin only Marlin has to recalculate the probing area.

If you have a 300x300 defined print surface and a 10mm probing margin, Marlin will want to probe a 280X280 square. One probing every 70mm (with 5x5 points probing).
As the endstops are active during bed leveling, Marlin will realize that it's not possible to reach X10 position (with 47mm probe X offset). He will then recalculate the probing area to go from the minimum physically possible probing position to the the set margin on the right of the print area. 
In the case of the CR-10 V3 that will give one probing every +/- 63mm measure when doing a real test with 10mm margin.

Marlin does : 
- Ok I'm trying to probe at X10 (with probe offset, So X-36 from nozzle point of view) but I'm reaching X Endstop before, so it means right now I'm at X-5 as defined with MIN_POS_X (nozzle position)
- I have a 46mm x probe Offset
- I means the min X position I can reach with the probe is : -5+46 = 41mm
- So my mesh will have to be from X41 to X290 (10mm margin)
- Divided by 4 (for five point) the result  is 62,25, strangely almost the same than what have been measured in real test.

With 45mm margin, it's giving a probe area about 210*210. Divided by 4 it's giving a theoric 60mm between probe points and it is what the machine exactly does when you try and measure. 

So with 10mm Margin : 
- The probing area is not centered to the print area.
- The center of the probing area is not aligned with the HOME probing position.
- The probing area is not square anymore
- The first Y probing line of the mesh will be 41mm from the edge of the print area when the last one will 10mm close to the edge.
- You are decreasing the probing density on the most used printing area. (It's not all days that you're printing pieces close to the print area dimensions).

You can see also that if you try to compile with the tramming wizard margin at 10 mm, like this : 
#define TRAMMING_POINT_XY { { 10, 10 }, { 290, 10 }, { 290, 290 }, { 10, 290 } }

You will get an error saying that those positions are not compatible with the defined probe offset : 

Marlin\src\gcode\bedlevel\../../feature/tramming.h:36:36: error: static assertion failed: TRAMMING_POINT_XY point 0 is not reachable with the default NOZZLE_TO_PROBE offset and PROBING_MARGIN.

So I'm still convinced those 45mm margin are correct and suitable to get something clean and centered.

config/examples/Creality/CR-10 V3/Configuration.h
PyreDev: I changed this back to commented as original.
 - You don't have a clear message that the print is waiting for heating before ABL
 - The heating is done after homing so I'm not sure it's good for accuracy if the mesh is populated from differences from homing values. Probably better to manually pre heat.

config/examples/Creality/CR-10 V3/Configuration.h
Vertabreak: you seem to be under the impression several settings work in a way that they do not.
furthermore the use of both corner leveling and the tramming wizard is semi redundant they are 2 features that do the same thing just with a different method.
given that marlin must have at least 1kb of free ram for proper functionality the addition of extraneous features is not advised.
GitHub really isn't the place for this type of discussion i would advise joining the marlin discord and discussing all of this with the devs & documentarians, that include my self. 
you seem to be under the impression your being helpful but due to your lack of experience and complete understanding, i need to inform you a number of these changes will ultimately lead to many bug reports.
your actions remind me of my self 5 years ago when i got my first 3dprinter diving in and playing with all the settings and options just in the control room throwing levers and pushing buttons with no real idea what anything did at the time. in time this fiddling with everything granted me the understanding of the nuances of marlin i have today.

PyreDev:
@AaronJust Hi, can you have a review of this update of the CR-10 V3 config you made ?

config/examples/Creality/CR-10 V3/Configuration.h
thinkyhead: With a +46mm probe X offset, the probe cannot reach the left side but it can definitely reach the right side, so you can set a very small margin, and only the left margin needs to be large. Give this a try:
```cpp
#define PROBING_MARGIN 10
#define PROBING_MARGIN_LEFT 45
```

config/examples/Creality/CR-10 V3/Configuration.h
thinkyhead: It may be someone's preference to set the bed center to 0, 0 for a Cartesian, so the option is exposed.

config/examples/Creality/CR-10 V3/Configuration.h
thinkyhead: If disconnecting the switch does nothing then the switch is likely a Normally Open (NO) switch. Filament runout is triggered only when the state _changes_ from HIGH to LOW.

thinkyhead:
> Probing area was not centered to Printing area

It's better to probe a maximum area than to probe a centered one.

config/examples/Creality/CR-10 V3/Configuration.h
thinkyhead: It should be based on having 0, 0 close to the front-left corner of the printable area, then adjust the bed size to achieve better centering.

config/examples/Creality/CR-10 V3/Configuration.h
thinkyhead: > You don't have a clear message that the print is waiting for heating before ABL

Everything is a work-in-progress. Please contribute to improve this facet!

PR #553, opened 2021-08-13
Tuned configuration for UMO+

ansonl:
### Description

Improved Marlin configuration for Ultimaker Original+ (board v2.1.1 and above)

### Benefits

Classic jerk enabled + Junction deviation disabled
Linear advance disabled until a correct Linear advance constant can be tested. 
Stepper motor current decreased from official UMO+ Marlin v1 values. 
Acceleration decreased from official UMO+ Marlin v1 values. 
Bed level corners enabled to replicate original UMO+ firmware bed leveling behavior with the paper test. 

Multiple Marlin and G-code commands enabled for enhanced UI and print functionality. 
Buffer size increased. 

CASE_LIGHT enabled using built in LED_PWM pin on Ultimain Board 2.1.1 (tested on 2.1.4). Recommended current is <250amps. 

### Related Issues

Improves upon UMO+ configuration from #510


config/examples/Ultimaker/Ultimaker Original Plus (2.1.1)/Configuration_adv.h
```diff
@@ -2156,7 +2156,7 @@
 
 // The ASCII buffer for serial input
 #define MAX_CMD_SIZE 96
-#define BUFSIZE 4
+#define BUFSIZE 16
```
thinkyhead, line 2167: Note that increasing the command buffer is usually pointless, and also causes pause to take longer because the buffer must reach the added `M25` command before pause is initiated.

PR #558, opened 2021-08-23
Update Longer3D/U20 examples

tpruvot:
- enable M575 serial baudrate
- Allow KILL command
- Handle the LK1+ like the U20+

### DIFF
--- a/config/examples/Alfawise/U20-bltouch/Configuration.h
+++ b/config/examples/Alfawise/U20-bltouch/Configuration.h
@@ -78,7 +78,20 @@
 //     U20_PLUS was not tested, as we do not have a printer to test. Print bed PID settings MUST be tuned for it.
 
 // Valid submodels: U20, U20_PLUS, U30, LK1, LK1_PLUS, LK2, and LK4
-// Valid touchscreens: TS_V11, TS_V12, and TS_V19 (2019)
+
+//#define U20
+//#define U20_PLUS
+//#define U30
+//#define LK1
+//#define LK1_PLUS
+//#define LK2
+//#define LK4
+
+// Valid touchscreens: TS_V11, TS_V12, and TS_V19 (2019). Select none for user-calibration.
+
+//#define TS_V11
+//#define TS_V12
+//#define TS_V19
 
 // 2 - Select the screen controller type. Most common is ILI9341 - First option. If your screen remains white,
 //     Try the alternate setting - this should enable ST7789V or ILI9328. For other LCDs... code is needed
--- a/config/examples/Alfawise/U20-bltouch/Configuration.h
+++ b/config/examples/Alfawise/U20-bltouch/Configuration.h
@@ -74,24 +74,11 @@
 // https://www.lesimprimantes3d.fr/forum/topic/18260-alfawise-u20x-u30-marlin-2x-firmware-alternatif/
 //
 // 1 - Select your Alfawise U30 or U20 or U20+ printer (NO MORE REQUIRED HERE)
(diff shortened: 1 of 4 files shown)


thinkyhead:
These don't seem to be defined anywhere. Commented-out defines would be helpful to provide context, either in the config or in the board's pins file.
```cpp
//#define U20
//#define U30
//#define U20_PLUS
//#define LK1
//#define LK1_PLUS
//#define LK2
//#define LK4
```

thinkyhead:
Well, defining just one of them in the `.ini` file will work, but are these defines needed for some external library? If not, they are fine to include in the example configuration, with an `#error` if none of them are defined. That also ensures Arduino IDE compatibility.

config/examples/Alfawise/U20/Configuration.h
```diff
@@ -75,12 +75,23 @@
...
 // 1 - Select your Alfawise U30 or U20 or U20+ printer (NO MORE REQUIRED HERE)
 //     and the touchscreen version, either 1.1 or 1.2. Most recent in France are v1.2 (blue pcb)
-//     These defines are set in platformio.ini build parameters, sample for U20 -DU20 -DTS_V12
 //     U20_PLUS was not tested, as we do not have a printer to test. Print bed PID settings MUST be tuned for it.
 
-// Valid platformio.ini submodel values are U20_PLUS U20 U30 LK1 LK2 LK4
+// Valid submodels: U20, U20_PLUS, U30, LK1, LK1_PLUS, LK2, and LK4
 
-// Valid platformio.ini touchscreens are TS_V11 TS_V12 TS_V19
+#define U20
```
tpruvot: yes but please do not define them here, its set per env to allow to generate multiple firmwares

config/examples/Alfawise/U20-bltouch/Configuration.h
```diff
@@ -2849,22 +2857,20 @@
...
-
-  #if ENABLED(TS_V19)
+  #elif ENABLED(TS_V19)
     // Longer LK4/U30 2.8" Ver 2019 / Blue PCB, SID240x320-8PCB-D
     #define TOUCH_CALIBRATION_X -12000
     #define TOUCH_CALIBRATION_Y  9000
     #define TOUCH_OFFSET_X       320
     #define TOUCH_OFFSET_Y       0
+  #else
+    #error "Please specify TS_V11, TS_V12, or TS_V19."
```
tpruvot: in theory this is not mandatory and could ask for user calibration... (TOUCH_SCREEN_CALIBRATION)

config/examples/Alfawise/U20/Configuration.h
thinkyhead: There's only a single environment in Marlin that pertains to this board, and it can only apply when building with PlatformIO and not other embedded tools like Arduino IDE. And, are you proposing to define 21 environments, one for each possible combination of printer model and screen?

thinkyhead: Truly, we don't want to ask users to edit `platformio.ini` as a routine step in building the firmware. It makes it harder for users to keep configurations for multiple printers, especially keeping them organized and up to date.

tpruvot: Yea, its fine commented, and there is no real need to create 11 env... basically just 3 are required for the 3 bed sizes. But Longer3D also has a git and use their model names/strings... Ideally the default env should be the smallest bed (U30) to avoid problems if they dont flash the right one... They also made a new tiny one named CUBE2 with this board...

PR #559, opened 2021-08-29
A20 series has noise on endstops this option is required

Vertabreak:
prevent issues is noise on endstops for the A20 series that is well known to have issues with out this option.


thinkyhead:
Interesting! I have never seen this issue on the A20M. But it can't hurt!

Vertabreak:
i had seen the problem on both my A20T V4+ board and a A20 with V4+ board. My V3 board models were not affected.
As you stated the option does not hurt anything even if not required so better off covering the bases.

PR #577, opened 2021-09-25
Update the thinker se and thinker v2 printer example configurations

Eryone:
### Requirements
None

### Description
We want to update the example configurations for the Thinker SE and Thinker V2 printer.

### Benefits
1. The stock Thinker SE and Thinker V2 printer don't come with Bltouch.
So we disable it as default. Now it is more friendly for the newcomer and player.
2. We have updated the configurations and it supports bltouch, filament runout sensor, and direct drive extruder system. We have moderately integrated these features. It means the user can choose the feature which they want to use.
3. We have updated the readme.md file of the configurations. The user can understand it
more clear now.
4. For the thinker v2 printer. We have enabled the "PIDTEMPBED" to improve the print quality result.
5. Bltouch requires the "z_min endstop logic to false". But the mechanical endstop logic is "true". So we have improved it and now the printer can work in both situations.

### Related Issues
None

### DIFF
--- a/config/examples/Eryone/Thinker SE/Configuration.h
+++ b/config/examples/Eryone/Thinker SE/Configuration.h
@@ -21,7 +21,9 @@
  */
 #pragma once
 
-#define THINKERV2_BLTOUCH  // Enable for an installed BLTOUCH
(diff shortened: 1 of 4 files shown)


Eryone:
Thanks!

PR #582, opened 2021-10-04
Disable BLTOUCH_SET_5V_MODE on Ender 3 / Ender 3 Max Creality boards

CRCinAU:
### Description

New  validation rules in pre-processing fail the build when `BLTOUCH_SET_5V_MODE` is set on these boards.

Update to disable `BLTOUCH_SET_5V_MODE` on these boards... There may be more, but these I know of for certain.


thisiskeithb:
Those aren’t STM32 configs, so how are they tripping the sanity checks?

CRCinAU:
The Ender 3 Max config notes its for a v4.2.2 board: https://github.com/MarlinFirmware/Configurations/blob/a489a3a4873a21e5a3195d5bae521a5d4ee346b0/config/examples/Creality/Ender-3%20Max/Configuration.h#L148

The Ender 3 has configs for both the v4.2.2 and v4.2.7 boards.

All are STM32 based.

The original 8 bit boards weren't STM32.

thisiskeithb:
> The original 8 bit boards weren't STM32.

Correct. I was looking in the wrong folders.

thisiskeithb:
I just ran a search and the following Creality configs should be updated as well:

- Ender-5 Pro/CrealityV422
- Ender-5 Pro/CrealityV427
- Ender-5/CrealityV427
- Ender-3 Pro/CrealityV427
- Ender-3/TH3D EZBoard Lite V2
- CR-10 S5/BigTreeTech SKR Mini E3 1.2 with TFT35 E3 V3.0
- CR-10 Mini/BigTreeTech SKR Mini E3 2.0

I *think* that's it.

CRCinAU:
> I _think_ that's it.

I've updated the ones you listed as well.

thisiskeithb (approved):

thisiskeithb:
Oh right. You have to target the `import` branch, not `bugfix`.

PR #596, opened 2021-10-22
update SDSORT_LIMIT back to 40 in CR-10/CrealityV1/Configuration_adv.h

ellensp:
### Requirements

Use example config from CR-10/CrealityV1

### Description

SDSORT_LIMIT of 256 exceeds available memory in the atmega2560.
Compiled firmware has 189% of ram used
This will not boot.

Set SDSORT_LIMIT back to 40

### Benefits

Firmware will not exceed available ram and will boot.

### Related Issues
Memory issue was noticed by Sound in Marlin Discord

### DIFF
--- a/config/examples/Creality/CR-10/CrealityV1/Configuration_adv.h
+++ b/config/examples/Creality/CR-10/CrealityV1/Configuration_adv.h
@@ -1433,7 +1433,7 @@
 
   // SD Card Sorting options
   #if ENABLED(SDCARD_SORT_ALPHA)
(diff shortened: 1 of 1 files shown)


thinkyhead:
As a Melzi board the `BOARD_MELZI_CREALITY` (1284P) actually has an excess of SRAM (16KB).

ellensp:
Sigh, user tricked me, swapped the controller with a BOARD_RAMPS_CREALITY and used CR-10/CrealityV1 Configs. 
It overflowed that 8k.
I didn't notice they had changed motherboards from a BOARD_MELZI_CREALITY

PR #605, opened 2021-11-01
Add Biqu B1 with SKR 2

thisiskeithb:
### Description

Due to chip shortages, the SKR 1.4 (and most other LPC-based boards) have been discontinued, so Biqu upgraded the B1's motherboard to an SKR 2.

Note: This config is not for the SE & SE Plus series which use a strain gauge as a probe by default.

### Benefits

SKR 2-based B1 owners can upgrade to the latest firmware.

### Related Issues

None. Reported by ROFLstomper in Biqu/BTT's Discord server.

edit: Thanks ROFLstomper for confirming these configs work!


thisiskeithb:
This still needs validating on real hardware (not sure if Biqu swapped anything else out), so I’ve marked it as draft.

thinkyhead:
To reduce the number of distinct configuration files I've combined BLTouch configs and non-BLTouch configs, with an added flag to enable BLTouch and its associated options.

PR #608, opened 2021-11-03
FFCP pins definition fix following PR22983

grauerfuchs:
### Description

PR 22983 on the Marlin bugfix-2.0.x repository removed the defines previously used for the FFCP extruder fan pins, thereby causing a failure to compile.

### Benefits

This change fixes the omission by directly assigning the pins used on the FFCP powered by the Rev. E (and Rev. D) Mightyboard. As the pins are still the same, the new config will compile against the code both prior to and post PR22983.


thisiskeithb:
I'd recommend `FAN1_PIN` and `CONTROLLER_FAN_PIN` instead:

https://github.com/MarlinFirmware/Marlin/blob/02b6fb8025ef82c2d1290cdc1a98bc0c54f502c9/Marlin/src/pins/mega/pins_MIGHTYBOARD_REVE.h#L185-L191:
```cpp
#ifndef FAN1_PIN
  #define FAN1_PIN                             7  // H4 EX1_FAN_PIN
#endif

#ifndef CONTROLLER_FAN_PIN
  #define CONTROLLER_FAN_PIN                  12  // B6 EX2_FAN_PIN
#endif
```

grauerfuchs:
The problem is, on the FFCP, they're really not either. Plus, both of those defines are based on logic that alters depending on various conditions and configurations. Therefore, it presents a situation that could in the future become an incorrect assignment whereas assigning the actual pin numbers would be stable. 

For instance: what if someone decides to switch one of the free pins via an external MOSFET board to drive a real controller fan, such as what is needed when swapping out the A4988s with TMC2208s? All of a sudden, that extruder's fan is broken. Given that it's an FFCP-specific thing versus generic use of the Mightyboard, it makes sense to update the FFCP-specific configuration example with something that isn't subject to change.

thisiskeithb:
Changes to the pins file should probably be reverted in the main Malrin repo.

grauerfuchs:
Again, the change I supplied in this PR is FFCP-specific and meant to apply to the FFCP configuration. The changes thinkyhead made in the main Marlin repo more closely follow the behavior of other boards when used in generic machines. Therefore, his logic is correct from a generic sense but incorrect only for the FFCP and equivalents.

grauerfuchs:
Why was this closed without action? The code in the bugfix branch of the main repo still will not compile with the configuration at present.

thisiskeithb:
> Why was this closed without action? The code in the bugfix branch of the main repo still will not compile with the configuration at present.

You have to target the `import-2.0.x` branch or PRs will be closed when the Configurations `bugfix-2.0.x` branch is froce-pushed since there is no longer a matching parent commit.

https://github.com/MarlinFirmware/Configurations/blob/073db3b25308907425cb6fe2fbb08fef04e879a5/README.md#L32

grauerfuchs:
Understood. Thank you. I'm not sure why I didn't see that.

thisiskeithb:
It was confusing when it happened to me as well 🙂

PR #632, opened 2021-12-13
for my cr-10

skylerjk4:
has slice engineering copperhead setting

### Requirements

* just changed my machines firmware


ellensp:
seems like this was meant to be for a private repo

PR #635, opened 2021-12-16
Update Configuration.h

rongoodman303:
Fix custom machine name to not be logic based, so ABM can display proper name. 
See ABM issue 51: (https://github.com/MarlinFirmware/AutoBuildMarlin/issues/51)

### Requirements
Autobuild Marlin must be able to display correct custom machine name, and cannot do so because there are currently multiple uncommented options surround by logic, which ABM cannot currently handle and parse. 

### Description

See ABM bug/enhancement #51

We must be able to understand your proposed change from this description. If we can't understand what the code will do from this description, the Pull Request may be closed at the maintainers' discretion. Keep in mind that the maintainer reviewing this PR may not be familiar with or have worked with the code recently, so please walk us through the concepts.

-->

### DIFF
--- a/config/examples/Alfawise/U20/Configuration.h
+++ b/config/examples/Alfawise/U20/Configuration.h
@@ -182,20 +182,22 @@
 
 // Name displayed in the LCD "Ready" message and Info menu
 //#define CUSTOM_MACHINE_NAME "3D Printer"
+// You must manually pick and uncomment one of the selections below so Autobuild Marlin
+// can display proper custom machine name.
 #if ENABLED(U20)
-  #define CUSTOM_MACHINE_NAME "Alfawise U20"
+//  #define CUSTOM_MACHINE_NAME "Alfawise U20"
 #elif ENABLED(U30)
-  #define CUSTOM_MACHINE_NAME "Alfawise U30"
+//  #define CUSTOM_MACHINE_NAME "Alfawise U30"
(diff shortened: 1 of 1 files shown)


thisiskeithb:
ABM can't parse many configs in this repo with printer-specific flags, so it'd be better to wait for true config parsing instead of "fixing" this single config.

thinkyhead:
> true config parsing

I'll work on that.

rongoodman303:
If the ABM issue gets escalated to a higher development priority, and will be resolved sooner than later, then I would agree with you. But ABM report https://github.com/MarlinFirmware/AutoBuildMarlin/issues/51 has been flagged as an enhancement, and could be far off. In the meantime, these kinds of issues in current example files makes them problematic for developers who need to use them. 

Until/if ABM is ever enhanced to better support logical selections of MOTHERBOARD and CUSTOM_MACHINE_NAME, whenever that may be, users need to be able to use these examples without these kinds of frustrating issues. When that happens, the logic in the examples can be restored/reverted, if necessary. That's why I did not propose removing the logic, just make it clear what the user/developer needs to do for now to let it work as expected.

rongoodman303:
Thanks. Just saw you commit to work on the ABM logic parsing issue. Thanks.

rongoodman303:
BTW: I would be happy to thoroughly test revised ABM when available. I have a Configuration.h already built that would be great for it :)

PR #637, opened 2021-12-25
Add Kingroon KP3S config

kubik369:
### Description

Adds a config for Kingroon KP3S (version 3.0 with the "Titan" extruder). I have taken the config from user [bdwilson](https://github.com/bdwilson/KP3S), played around with it and brought it up to date with 2.0.9.3.

Some remarks:
- I am not (yet) knowledgeable enough regarding the acceleration values and other such variables, so I went with the values from bdwilson's config . For what it's worth, the few prints I tried look pretty much perfect to my untrained eye.
- I am not sure what is the policy regarding the enabling of features in these configs. I was not able to find what is the "standard" for these configs, so if some things need to be disabled to conform, please let me know.

Regarding some config choices:
- Kingroon KP3 v1.2 board which is apparently based on MKS Robin Nano.
- Marketed 180x180x180 mm bed, realistically you can get 165x170x190-200mm. I have personally set height to 190 and will try 200 when I get the courage to test it. The Z profile seems to be high enough, but probably better to have a safer config with the marketed 180mm. X is moved to the right (because of fixed endstop switch) and shorter because of the way Y rail is mounted and because of the belts mounted by two zipties, which blocks like 10mm from one side. Also some people have the bed crooked, as in a small rotation in the XY plane due to the way how the bed is mounted to the carriage which basically shortens the usable Y range. This can probably be easily fixed, but again, probably a safer config is better.
- The advanced config should be identical to the one from the example folder.

Thanks for the review :)


Skanderberg777:
Hi there,

if I download the configurations for the new Marlin, the Kingroon folder just has a KP§ folder, no KP3S. 

Opened in Vistual Studio, the printer name is just KP3.

Where do I find the KP3S?

Thank you very much.

thinkyhead:
https://raw.githubusercontent.com/kubik369/Configurations/957830f430cf80fc79843c7a36f29c1b750aee0b/config/examples/Kingroon/KP3S/Configuration.h

yla199503:
Hi there,
What should I do if anyone has a Kingroon KP3 v1.3 board?
and nope Configuration_adv.h

thisiskeithb:
@yla199503 

This Issue Queue is for Marlin bug reports and development-related issues, and we prefer not to handle user-support questions here. (As noted on [this page](https://github.com/MarlinFirmware/Marlin/blob/bugfix-2.1.x/.github/contributing.md#i-dont-want-to-read-this-whole-thing-i-just-have-a-question).) For best results getting help with configuration and troubleshooting, please use the following resources:

- [Marlin Documentation](https://marlinfw.org) hosted on MarlinFW.org
- [Marlin Forum](https://forums.reprap.org/list.php?415) hosted on RepRap.org
- [Marlin Firmware](https://www.facebook.com/groups/1049718498464482/) Facebook Group 
- [Marlin Firmware for 3D Printers](https://www.facebook.com/groups/3Dtechtalk/) Facebook Group
- [Marlin Configuration](https://www.youtube.com/results?search_query=marlin+configuration) on YouTube
- [Marlin Firmware](https://discord.gg/n5NJ59y) Discord server

PR #648, opened 2022-01-03
Added Elegoo Neptune 2 board configurations

just-trey:
### Requirements

### Description

Added some example configurations for The Elegoo Neptune 2 using  `TFT_COLOR_UI`. The Elegoo Neptune 2 can come with a ZNP Robin Nano 1.2 or 1.3 Board (The are MKS Clone boards). I have created configurations for both versions of the machine. and also have a configuration for the bl touch for each version

### Benefits

Adde examples of how to build configurations for Elegoo boards

### Related Issues

I would  have requested to remove this file: https://github.com/MarlinFirmware/Configurations/tree/import-2.0.x/config/examples/Elegoo/Neptune%202 but I do not see it in the import branch
I have tested this configuration and it does not work and also causes issues with PID tuning from what I can tell


thinkyhead:
Please reconcile this submission with #636 by @EvilGremlin

EvilGremlin:
_ENDSTOP_INVERTING true ? machine revisions are seriously that different?
Default fan PWM works for you?

EvilGremlin:
Checked my config with 2.0.9.3, nothing wrong with PID, autotune works, heatup swing within 5C - no problem here. Didn't check bed though, but bed values are not critical anyway.
image

EvilGremlin:
Trey, so what were actual problems with my config not working for your machine? Especially, what the hell going on with endstops?

just-trey:
@EvilGremlin, First thanks, I really appreciate the configuration you wrote.

I was confirming what you had set as numbers in your config vs. what Elegoo has as the official version of their firmware on their site (Marlin 1 with LVGL enabled). In their .txt configuration, this is what they have and it is consistent across all versions of their firmware.

```
# Mechanical endstop with COM to ground and NC to Signal uses "false" here (most common setup).
>X_MIN_ENDSTOP_INVERTING 0		# X_MIN limit switch normally open/normally closed selection
>Y_MIN_ENDSTOP_INVERTING 0		# Y_MIN limit switch normally open/normally closed selection
>Z_MIN_ENDSTOP_INVERTING 0		# Z_MIN limit switch normally open/normally closed selection
>X_MAX_ENDSTOP_INVERTING 0		# set to true to invert the logic of the endstop.
>Y_MAX_ENDSTOP_INVERTING 1		# set to true to invert the logic of the endstop.
>Z_MAX_ENDSTOP_INVERTING 1		# Z_MAX limit switch normally open/normally closed selection
>Z_MIN_PROBE_ENDSTOP_INVERTING       1	# Z_PROBE limit switch normally open/normally closed setting.
```
I did make a change now after a quick check and this is what I think is correct

```
#if ENABLED(HAS_BLTOUCH)
  #define Z_MIN_PROBE_ENDSTOP_INVERTING false // Set to true to invert the logic of the probe.
#else
  #define Z_MIN_PROBE_ENDSTOP_INVERTING true // Set to true to invert the logic of the probe.
#endif
```

If I am misreading something please let me know. The main issue I ran into is the one you have on the site will not load on the 1,.3 board that I have (You may have the 1.2 board). I also have enabled more features and it's not up yet, but I have gotten down to one config where everything is configurable at the top of the file. It's not quite ready for PR as I am still testing but this is my latest here: https://github.com/just-trey/Marlin/tree/elegoo-neptune-2/config/Elegoo/Neptune-2. I am a big fan of config switches and it may be overkill but you can configure almost everything here to manage the board version and if you have a bltouch from here. https://github.com/just-trey/Marlin/blob/elegoo-neptune-2/config/Elegoo/Neptune-2/Configuration.h#L80-L82

I am adding 2d configuration right now so more things are on their way in. 

My readme and version files are not updated yet.

EvilGremlin:
Yep, mine is 1.2. But i think i never actually looked at their configs... so they changed board and endstops then, and no way to know where's what without some digging... oh wait, y_max and z_max? then there's no conflict, i only have min endstops. This is their mess that we ignore.

just-trey:
yeah, I'm just being a bit OCD. Onec I have my configs up I'll be sure to let you know. Probably sometime tomorrow later. I have to run some benchmark prints to make sure I did not break anything.

just-trey:
@thinkyhead  and @EvilGremlin  this is ready for re-review. Thanks!

config/examples/Creality/Ender-3/BigTreeTech SKR Mini E3 2.0/Configuration.h
```diff
@@ -1611,7 +1611,7 @@
   /**
    * Enable the G26 Mesh Validation Pattern tool.
    */
-  //#define G26_MESH_VALIDATION
+  #define G26_MESH_VALIDATION
```
thisiskeithb: This seems like an unrelated change.

EvilGremlin: yep, generated pattern is borderline useless...

config/examples/Elegoo/Neptune-2/Configuration.h
```diff
@@ -1627,7 +1655,7 @@
 //#define AUTO_BED_LEVELING_3POINT
 //#define AUTO_BED_LEVELING_LINEAR
 //#define AUTO_BED_LEVELING_UBL
-#if ENABLED(HAS_BLTOUCH)
+#if HAS_BLTOUCH
   #define AUTO_BED_LEVELING_BILINEAR
 #else
   #define MESH_BED_LEVELING
```
EvilGremlin: Why not UBL for both? It's better to leave leveling types all disabled, because it's mostly a matter of taste.

config/examples/Elegoo/Neptune-2/Configuration_adv.h
```diff
@@ -2337,7 +2339,7 @@
 //#define NO_TIMEOUTS 1000 // Milliseconds
 
 // Some clients will have this feature soon. This could make the NO_TIMEOUTS unnecessary.
-#define ADVANCED_OK
+//#define ADVANCED_OK
```
EvilGremlin: These are all for proper serial communucation, leaving it on default is unwarranted IMO, it'll just increase amount of questions "why octoprint lose connection?"

EvilGremlin:
About merged config for 2 machines... how'd it work with AutoBuild?

config/examples/Creality/Ender-3/BigTreeTech SKR Mini E3 2.0/Configuration.h
just-trey: turned back off in the latest code.

config/examples/Elegoo/Neptune-2/Configuration.h
just-trey: @EvilGremlin ,I see your point. I got this idea from looking at config for a machine with a similar board setup. [Two Trees - BlueR V3](https://github.com/just-trey/Configurations/blob/import-2.0.x/config/examples/Two%20Trees/BlueR/BlueR%20V3). My opinion is we leave it at is or leave leveling types all disabled. If someone wants to use UBL they would understand this enough to make the changes. Thoughts?

config/examples/Elegoo/Neptune-2/Configuration_adv.h
just-trey: reverted in the latest. Thanks for helping me understand better what this does.

just-trey:
> About merged config for 2 machines... how'd it work with AutoBuild?

Not very well. it would get "stuck" on whatever version you built on. I could not get it to refresh the board type in auto build. I am not sure what happened. At one point it was stuck on the 1.2 board config. Then it changed to the 1.3 board config at some point and won't go back to 1.2 for me. I ended up using the pio clean and building steps to get around it. We could have 5 or 6 separate files. I was just a fan of the switches. Would like to hear your thoughts on this.

thinkyhead:
Auto-build is not compatible with overly-complicated configurations. It doesn't interpret the configs, it just scrapes them.

config/examples/Elegoo/Neptune 2/Configuration.h
```diff
@@ -69,9 +69,60 @@
 // @section info
 
 // Author info of this build printed to the host during boot and M115
-#define STRING_CONFIG_H_AUTHOR "(EvilGremlin)" // Who made the changes.
+#define STRING_CONFIG_H_AUTHOR "(just-trey, mlee12382, and EvilGremlin, Neptune 2 - Marlin 2 Color UI)" // Who made the changes.
```
thinkyhead: A string this long may be problematic. Watch carefully for crashes and buffer overflows.

just-trey (commented):
@thinkyhead updated author string

Issue #644, opened 2021-12-29
[FR] Ender 3 Pro with SKR mini E3 V3.0


HoreaM:
### Description

I searched in the configurations folder for the Ender 3 Pro, but it seems like the only available configuration files are for the SKR mini E3 v1/1.2/2 boards. It would be of great help if we could get an "official" configuration file for the Ender 3 Pro with the new SKR mini E3 V3.0 board, as I'm not entirely sure what needs to be adjusted from the previous SKR mini E3 versions. Thank you!

### Configuration Details

* Manufacturer: Creality
* Model: Ender 3 Pro
* Motherboard: SKR mini E3 V3.0

HoreaM:
I just noticed that the Ender 3 has a configuration file for the SKR mini E3 V3.0. Could I use that one for my Ender 3 Pro? Thank you!

thisiskeithb:
> Could I use that one for my Ender 3 Pro?

Yes. As far as Marlin is concerned, the only difference is the printer name and status icon text.

thisiskeithb:
Having said that, here's a config: [Ender-3 Pro - BigTreeTech SKR Mini E3 3.0.zip](https://github.com/MarlinFirmware/Configurations/files/7875628/Ender-3.Pro.-.BigTreeTech.SKR.Mini.E3.3.0.zip)

~~I'll put in a PR later.~~ Done!

thisiskeithb:
Added in https://github.com/MarlinFirmware/Configurations/pull/660

PR #660, opened 2022-01-15
Add Ender-3 Pro with SKR Mini E3 V3

thisiskeithb:
### Description

Add Ender-3 Pro with SKR Mini E3 V3.

### Related Issues

- #644


LazeMSS:
@thisiskeithb just wondering (was doing the profile for my Ender 3 pro myself) and did a compare with the only public source code from Creality (for Marlin 1.x) - i see a lot of difference in the values for that and many of the other Creality examples in here.

Examples:
#define HOMING_BUMP_DIVISOR {2, 2, 4} in creality original source - in almost every example in here its HOMING_BUMP_MM      { 5, 5, 2 } and I have seen configurations where its even { 0, 0, 2 }

I have attached my WIP for the Ender SKR mini v3.0 with all fans connected and the pinout naming is used from the pin files instead of the direct pin assignment (FAN2_PIN instead of PB15)

Fan assigment:
Part cooling fan: FAN0_PIN
Hotend fan: FAN1_PIN (will turn fan on at extruder temp = 50)
Motherboard/stepper fan: FAN2_PIN

It also includes BLTouch i was wondering it there should be an example with that included or not? And all the options to make OctoPrint more usefull.

[skr-mini-v3-config.zip](https://github.com/MarlinFirmware/Configurations/files/7892066/skr-mini-v3-config.zip)

thisiskeithb:
I would ignore anything from Creality at this point, but I did plan to go through the Ender-3/5 configs and bring them closer to the standard (keeping improvements for 32-bit boards enabled) since they're all over the place right now.

> It also includes BLTouch i was wondering it there should be an example with that included or not? 

Based on previous guidance, the plan is to remove all the extra BLTouch configs and merge them into the base printer config with a switch like `PRINTER_NAME_WITH_BLTOUCH`.

LazeMSS:
Awesome. Yes someone the profiles are really a bit off.

How about using the aliases for the PIN numbers instead of the actual pin names?

thisiskeithb:
> How about using the aliases for the PIN numbers instead of the actual pin names?

I can do that (and generally try to), but I knew I was going to overhaul bunch of configs soon anyway 🙂

~~I’m mobile right now, but I’ll push an update later.~~ Done.

config/examples/Creality/Ender-3/BigTreeTech SKR Mini E3 3.0/Configuration_adv.h
```diff
@@ -512,7 +512,7 @@
  */
 #define USE_CONTROLLER_FAN
 #if ENABLED(USE_CONTROLLER_FAN)
-  #define CONTROLLER_FAN_PIN PB15           // Set a custom pin for the controller fan
+  #define CONTROLLER_FAN_PIN FAN2_PIN       // Set a custom pin for the controller fan
```
thinkyhead, line 515: There is a small risk here, if in future we decide to `#undef` pins that are not actually used. But for now this is safe, since we check for `CONTROLLER_FAN_PIN` being connected to a board fan plug in `Conditionals_post.h`.

config/examples/JGAurora/A1/Configuration_adv.h
```diff
@@ -116,7 +116,7 @@
 #define HEATER_BED_INVERTING 1
 #define HEATER_0_INVERTING 1
 //#define HEATER_1_INVERTING 0
-#define FAN_INVERTING 1
+#define FAN_ACTIVE_STATE LOW
```
thinkyhead: I have a pending PR for this change, but right now it's too soon. At the moment I forget which PR this change is associated with.

config/examples/Creality/Ender-3/BigTreeTech SKR Mini E3 3.0/Configuration_adv.h
thisiskeithb: These fan* pin aliases make it easier to move a config between boards / upgrades, so it’d be nice if they can stick around.

config/examples/JGAurora/A1/Configuration_adv.h
thisiskeithb: That wasn’t my commit, so please change what you need to.

The original commits should be good to go.

config/examples/Creality/Ender-3/BigTreeTech SKR Mini E3 3.0/Configuration_adv.h
thinkyhead: Yeah, generally it's nice to have pin defines based on the labeled pins on the board, but it quickly gets ugly if we try to do this for all pins on all boards, so this is an ok middle ground.
- name pins for their expected uses
- keep them around for other uses (avoid `#undef`ing)
- not assume a function is active just because a pin for it is defined

We used to avoid checking for enabled features in `pins.h` but now it's basic practice, and that's fine with me!

PR #685, opened 2022-02-27
add support for Weedo TINA2 printer

atsju:
### Requirements

https://github.com/MarlinFirmware/Marlin/pull/23817

### Description

Add support for a market printer [WEEDO TINA2](http://www.weedo.ltd/product-detail/tina2/)
Please note I am not associated to WEEDO in any way, only a private contributor.
I based my work on their [repo](https://github.com/weedo3d/TINA2firmware) which is based on Marlin.
Please confirm that the GPL licence makes this addition possible without Weedo acknowledgement.

Add did not add the custom bootscreen. Let me know what is prefered in this kind of cases.

### Benefits
 add a new printer support

### Related Issues
None

### DIFF
--- a/config/examples/Weedo/Tina2/V2/Configuration.h
+++ b/config/examples/Weedo/Tina2/V2/Configuration.h
@@ -1,6 +1,6 @@
 /**
  * Marlin 3D Printer Firmware
- * Copyright (c) 2020 MarlinFirmware [https://github.com/MarlinFirmware/Marlin]
+ * Copyright (c) 2022 MarlinFirmware [https://github.com/MarlinFirmware/Marlin]
  *
  * Based on Sprinter and grbl.
  * Copyright (c) 2011 Camiel Gubbels / Erik van der Zalm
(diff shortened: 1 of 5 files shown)


atsju:
I forgot to add I only own the basic version without wifi. So I'm not able to test the config file needed for Wifi support.

config/examples/Weedo/Tina2/V2/Configuration.h
```diff
@@ -1912,7 +1912,7 @@
 
 #if ENABLED(NOZZLE_PARK_FEATURE)
   // Specify a park position as { X, Y, Z_raise }
-  #define NOZZLE_PARK_POINT { (X_MAX_POS - 10), (Y_MAX_POS - 10), 20 }
+  #define NOZZLE_PARK_POINT { (X_MAX_POS - 10), (Y_MAX_POS - 10), (Z_MAX_POS - 10) }
```
thisiskeithb, line 1916: If you’re printing near max height and cancel, this could potentially damage hardware.

atsju: I though `G27` does `G27 P0` by default then it only goes up if not already higher. 
Please confirm

atsju: Anyway, feel free to change to `Z_MAX_POS` if you think it's necessary

thinkyhead: The value is a raise, not a coordinate (`G27` modes `P0` and `P2`). Does the nozzle really need to raise all the way to the top on any pause? Seems to be a small printer, so at least it would not take too long.

xxoconenrxx:
Good day I have the tina 2 verison now tested problem is the axes can not be set via the display z-offset so unfortunately not possible baby steps are also not possible do you have a solution?

PR #691, opened 2022-03-02
Biqu Thunder Standard Config

thisiskeithb:
### Description

Add the *discontinued* Biqu Thunder Standard config.

There's an "Advanced" version with TMC2208s and additional hardware, but that firmware was not posted & I don't have a way to easily verify settings.


thisiskeithb:
I need to check on some things and hopefully get some Thunder users to test it since they’ve been asking, so I’ll close this PR for now.

PR #700, opened 2022-03-21
Enable interrupts on SKR Mini E3 3.0

CRCinAU:
The SKR Mini E3 V3.0 mainboard has issues detecting triggers from a
BLTouch when polling endstops instead of using interrupts.

Enable interrupts on the default configs, and enable software mode in
the BLTouch so we get a longer pulse and even less chance of failure.

### DIFF
--- a/config/examples/Creality/CR-10 S5/BigTreeTech SKR Mini E3 1.2 with TFT35 E3 V3.0/Configuration.h
+++ b/config/examples/Creality/CR-10 S5/BigTreeTech SKR Mini E3 1.2 with TFT35 E3 V3.0/Configuration.h
@@ -893,7 +893,7 @@
 
 // Enable this feature if all enabled endstop pins are interrupt-capable.
 // This will remove the need to poll the interrupt pins, saving many CPU cycles.
-//#define ENDSTOP_INTERRUPTS_FEATURE
+#define ENDSTOP_INTERRUPTS_FEATURE
 
 /**
  * Endstop Noise Threshold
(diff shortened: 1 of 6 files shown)


config/examples/Creality/CR-10 S5/BigTreeTech SKR Mini E3 1.2 with TFT35 E3 V3.0/Configuration.h
```diff
@@ -893,7 +893,7 @@
 
 // Enable this feature if all enabled endstop pins are interrupt-capable.
 // This will remove the need to poll the interrupt pins, saving many CPU cycles.
-//#define ENDSTOP_INTERRUPTS_FEATURE
+#define ENDSTOP_INTERRUPTS_FEATURE
```
thisiskeithb, line 896: I’m sure it’s fine, but this is the SKR Mini E3 V1.2 config, not the V3.

CRCinAU: Yeah - @ellensp and I had a chat after (it was a mistake), but I do believe the v1.2 and v2.0 boards have full support for interrupts on the endstop / probe pins - so I decided to not correct it...

It may well be useful to enable this option across all of the SKR Mini E3 series of boards...

rq3: > I’m sure it’s fine, but this is the SKR Mini E3 V1.2 config, not the V3.

A question. Is there a reason why interrupts are NOT enabled by default in all configurations? From what I can see from data sheet review, if a pin or processor is incapable of accepting the request, it will be ignored?

On the other hand, enabling interrupt capability can, as the OP suggested, make a very impressive improvement in response time for those processors capable of actually doing it..

ellensp:
@rq3 If the pin cant do interrupts marlin it errors out,
eg
`      static_assert(digitalPinHasPCICR(X_MAX_PIN), "X_MAX_PIN is not interrupt-capable. Disable ENDSTOP_INTERRUPTS_FEATURE to continue.");
`
As it should

PR #707, opened 2022-04-06
Add missing Configuration_adv.h files

thisiskeithb:
### Description

Add missing `Configuration_adv.h` files


thinkyhead:
If a `Configuration_adv.h` file is identical to the default, then it can be omitted.

thisiskeithb:
> If a `Configuration_adv.h` file is identical to the default, then it can be omitted.

These are & there are probably quite a few that can be purged from this repo if that's the case. I'll do some comparing later and re-submit with those removed.

Issue #402, opened 2021-01-05
SD card reader disabled on Creality CR-10S


djardi:
### Bug Description

The most recent Marlin firmware update disables the SD card reader in Creality CR-10S. This is a configuration issue, not a bug. Having said that, since CR10-S printers come with an SD card slot and reader and since the proper configuration of this setting is rather obscure, I propose that the CR10-S example configuration files should be changed so that the SD card reader is enabled by default. That will save many users a lot of time and frustration trying to use their SD card reader after the update. This is easily accomplished by changing one line of the settings in the configuration_adv.h file:
change
//#define SDCARD_CONNECTION LCD
to 
#define SDCARD_CONNECTION ONBOARD

### Configuration Files
[SD_prob.zip](https://github.com/MarlinFirmware/Marlin/files/5767194/SD_prob.zip)

If you've made any other modifications describe them in detail here.

### Steps to Reproduce

1. [First Step]: Download the Marlin 2.0.x firmware for the Creality CR-10S (Configurations-release-2.0.7.2\Configurations-release-2.0.7.2\config\examples\Creality\CR-10S\CrealityV1). Use the configuration files found in the directory. Do NOT change the settings for the SD drive in the configuration_adv.h file.
2. [Second Step]: Try to use the SD card reader. The only menu choice related to the SD card reader is "Attach media," which does nothing. Restarting the printer does not help.

**Expected behavior:**
The menu should show the correct menu options and allow the user to print from the SD card.
![3](https://user-images.githubusercontent.com/76797450/103609076-df837d80-4ed1-11eb-934e-b648144fb450.jpg)
![2](https://user-images.githubusercontent.com/76797450/103609067-d8f50600-4ed1-11eb-9243-6a3a53b94a71.jpg)

**Actual behavior:**
There is only one menu choice (Attach media). Selecting this choice does not do anything with the SD card because the SD card reader has not been defined in the setup files. It does not matter whether the printer is turned on with the SD card in the slot or whether the SD card is inserted after the printer is turned on. The printer does not see the SD card. The user is unable to see any files on the SD card. Consequently, printing from the SD card is impossible.
![1](https://user-images.githubusercontent.com/76797450/103609087-ea3e1280-4ed1-11eb-9a68-423a391ec84a.jpg)

This is NOT a bug. It is a configuration issue. It seems to me to make more sense to have the default configuration of the configuration_adv.h menu set so that the default SD card reader in CR 10S printers is immediately available to users.

[...]

PR #716, opened 2022-04-20
Enable CR10S Onboard SD Card Reader

thisiskeithb:
### Description

Enable CR10S onboard SD card reader.

### Benefits

Onboard SD card reader works again.

### Related Issues

- https://github.com/MarlinFirmware/Configurations/issues/402

### DIFF
--- a/config/examples/Creality/CR-10S/CrealityV1/Configuration_adv.h
+++ b/config/examples/Creality/CR-10S/CrealityV1/Configuration_adv.h
@@ -1649,7 +1649,7 @@
    *
    * :[ 'LCD', 'ONBOARD', 'CUSTOM_CABLE' ]
    */
-  //#define SDCARD_CONNECTION LCD
+  #define SDCARD_CONNECTION ONBOARD
 
   // Enable if SD detect is rendered useless (e.g., by using an SD extender)
   //#define NO_SD_DETECT


thisiskeithb:
@thinkyhead: In https://github.com/MarlinFirmware/Marlin/issues/23518#issuecomment-1105652946, you mention that:
> `SDCARD_CONNECTION` doesn't apply to AVR boards.

I wonder why the UI / menu is different when `SDCARD_CONNECTION` enabled & set to `ONBOARD` as reported & shown in https://github.com/MarlinFirmware/Configurations/issues/402.

thinkyhead:
> I wonder why the UI / menu is different when `SDCARD_CONNECTION` enabled & set to `ONBOARD` as reported & shown in #402.

They're looking at 2.0.7.2 and I don't think we have the same issue today.

thinkyhead:
I see… The Creality RAMPS board is a special case.
```cpp
#ifndef SD_DETECT_PIN
  #if SD_CONNECTION_IS(ONBOARD)
    //#define HAS_ONBOARD_SD_DETECT               // If the SD_DETECT_PIN is wired up
  #endif
  #if ENABLED(HAS_ONBOARD_SD_DETECT) || !SD_CONNECTION_IS(ONBOARD)
    #define SD_DETECT_PIN                     49
  #endif
#endif
```

thinkyhead:
The change was made in MarlinFirmware/Marlin#23740 by @ellensp so we'll need to look at the reason it was requested.

thisiskeithb:
Oh, oops. I closed this too early then.

thinkyhead:
In `pins_RAMPS.h` the `SD_DETECT_PIN` will end up being 49 (`EXP2_04_PIN`==`AUX3_02_PIN`) if there is an RRD LCD or `CR10_STOCKDISPLAY`. Without the common LCDs it is simply not assigned. The change proposed in this PR should not make any difference, except to prevent `SD_DETECT_PIN` being assigned the value 49 in the block shown above.

thinkyhead:
The standard Creality3D V1.1.2 board has an onboard SD card and a single 10-pin connector for a stock display. There is no SD card on the stock display itself. However, `pins_RAMPS.h` still picks pin the same pin, 49, for any RAMPS that has that display, if no other value was previously given to `SD_DETECT_PIN`.

Anyway, it looks like what users might really want to do is enable `HAS_ONBOARD_SD_DETECT` to get the desired result, which then applies no matter which display is attached.

I'm not sure how you get an external SD card with the single 10-pin connector designed for the stock display, but maybe there are such units out there.

thinkyhead:
I guess the "onboard" idiom is fine, and maybe there are some other AVR boards with built-in SD that could also use external SD, so we'll just patch them as we find them.

InsanityAutomation:
> The standard Creality3D V1.1.2 board has an onboard SD card and a single 10-pin connector for a stock display. There is no SD card on the stock display itself. However, `pins_RAMPS.h` still picks pin the same pin, 49, for any RAMPS that has that display, if no other value was previously given to `SD_DETECT_PIN`.
> 
> Anyway, it looks like what users might really want to do is enable `HAS_ONBOARD_SD_DETECT` to get the desired result, which then applies no matter which display is attached.
> 
> I'm not sure how you get an external SD card with the single 10-pin connector designed for the stock display, but maybe there are such units out there.

The Creality ramps is the 2.2 with the 2 connector display. The older 1.1.4 etc board is a melzi / sanguino.

Issue #713, opened 2022-04-13
[BUG] Sunlu S8 configuration does not enable hotend heat sink fan.


stefanhra:
### Bug Description

The Sunlu S8 printer uses a PWM pin to control the hotend heat sink fan which turns on and off based on extruder temperature. The configuration files provided here do not enable this PWM control for the heat sink fan so it never turns on. The configuration.h file provided by the manufacturer for Marlin 1.x uses MOTHERBOARD BOARD_MKS_BASE while the configuration.h here uses MOTHERBOARD MOTHERBOARD BOARD_RAMPS_14_EFB.

### Configuration Files

**Required:** Include a ZIP file containing `Configuration.h` and `Configuration_adv.h`.

If you've made any other modifications describe them in detail here.

### Steps to Reproduce

1. Upload Marlin firmware to Sunlu S8 printer using configuration files from this project.
2. Set temperature of hotend to printing temperature (eg. 220 C)
3. Observe hotend heat sink fan. 

**Expected behavior:**

Hot end heat sink fan should start running when hot end heats up. 

**Actual behavior:**

Hot end heat sink fan doesn't run. 

#### Additional Information

* Provide pictures or links to videos that clearly demonstrate the issue.
* See [Contributing to Marlin](https://github.com/MarlinFirmware/Marlin/blob/2.0.x/.github/contributing.md) for additional guidelines.
[Sunlu S8.zip](https://github.com/MarlinFirmware/Configurations/files/8484832/Sunlu.S8.zip)

thisiskeithb:
According to @InsanityAutomation https://github.com/MarlinFirmware/Configurations/pull/512#issuecomment-1101377824, Sunlu must have shipped these out with RAMPS boards and MKS Base boards, so I'll create a new config.

InsanityAutomation:
The board change alone does not define an auto fan pin. The MKS base simply sets the board name then includes the ramps file, so functionally irrelevant. The auto fan pin would be defined in config adv. I did a quick bit of google fu, and it seems the config published has this line added : 
#define E0_AUTO_FAN_PIN 7

Which on machines newer than mine which was wired right to power would add the functionality you described.

thisiskeithb:
> The board change alone does not define an auto fan pin. The MKS base simply sets the board name then includes the ramps file, so functionally irrelevant.

I’m well aware of this. It’s not my first config 😛 

[...]

PR #718, opened 2022-04-21
Enable Sunlu S8 Auto Fan

thisiskeithb:
### Description

Enable Sunlu S8 auto fan.

The config really needs a new motherboard added to Marlin (as `SH_2560_A4988_V1`) instead of using the generic RAMPS motherboard define, but this will work for now:

### Benefits

Auto fan will work.

### Related Issues

- #713


InsanityAutomation:
Just as a quick note, early machines have the hotend fan wired to always on power, so while this is not needed it safely toggles the pin thats not landed to anything in the background. This allows both machines revisions to work under the same config.

InsanityAutomation (approved):

PR #725, opened 2022-05-01
K8200 Upgraded Z axis and full graphic controller

pau1ie:
### Requirements

Velleman K8200 with various options.

### Description

I have a Velleman K8200 with the VM8203 upgraded Z axis and full graphic controller, so I thought I would extend the config to allow those two options.

### Benefits

Previously the config only allowed for the VM8201 two line display option. Now the config can handle the following options

**Display**
- VM8201 Display
- Full Graphic Controller
- None

**Z Axis**
- VM8203 upgraded Z axis
- Original Z axis

### Related Issues

None.

I can't seem to test the config as is, so have done my best to create a hopefully working config. I took the changes from my working config and applied it to this file. I also can't test the non-upgraded Z axis or the VM8201 as I don't have those. I may have broken those, but I have tried to be careful.


thisiskeithb:
> I can't seem to test the config as is, so have done my best to create a hopefully working config. I took the changes from my working config and applied it to this file. I also can't test the non-upgraded Z axis or the VM8201 as I don't have those. I may have broken those, but I have tried to be careful.

If changes are that disruptive to a stock machine (and can't be verified/tested), I'd suggest creating a new config with a readme that explains all the upgrades so it's clear to users down the road.

PR #729, opened 2022-05-06
Elegoo neptune 2 max temp update

just-trey:
### Requirements

### Description

Reset max hotend temp to manufacturer's specifications as suggested by @tinkyhead

### Benefits

revert back to the manufacturer stock settings as not everyone has an all-metal hotend


thinkyhead:
Since the max temperatures are intended to include the overshoot, the hotends add 15 (275) and bed adds 10 (120). So the existing values might in fact be correct, unless the manufacturer actually intends constant temperatures of 245 and 100, which would be reasonable for PLA, PETG, ABS, etc.

just-trey:
@thinkyhead I do beleive the numbers here do meet the expectation of the manufacturer. This is an ender 3 clone and does not come with an all metal hotend. These temps seem safer to me after a review.

PR #738, opened 2022-05-17
upgrade ender 5 plus config

alex255:
### Description

upgraded and tested config for ender 5 plus and enable the new and usefull features of the Marlin 2
this config is created for Marlin 2.0.9.3

### Benefits

- set max hotend temp to 300 because the base hotend can handle this without problem.
- enable and config PIDTEMPBED
- increase max extrude lenght for longer bowden and filament change
- modify Z offset to correct value
- slow down the leveling and home Z speed ( the default is too fast for the machine )
- enable Z_MIN_PROBE_REPEATABILITY_TEST
- enable FILAMENT_RUNOUT_SCRIPT
- always enable auto leveling after home
- enable PREHEAT_BEFORE_LEVELING
- enable HOTEND_IDLE_TIMEOUT
- modify unload/load lenght to the real bowden lenght
- modify ADVANCED_PAUSE_FEATURE settings, and now work perfectly with and without screen (octoprint) 
- enable more M114 report options for octoprint

### DIFF
--- a/config/examples/Creality/Ender-5 Plus/Configuration.h
+++ b/config/examples/Creality/Ender-5 Plus/Configuration.h
@@ -20,8 +20,8 @@
  *
  */
 #pragma once
-#error "Don't build with import-2.0.x configurations!"
-#error "Use the 'bugfix...' or 'release...' configurations matching your Marlin version."
+
+#define CONFIG_EXAMPLES_DIR "Creality/Ender-5 Plus"
 
 /**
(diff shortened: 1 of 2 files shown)


thisiskeithb:
You’ve overwritten an `import-2.0.x` config with one from `2.0.9.3`.

Please start over with a fresh copy of `import-2.0.x` and implement your changes again.

PR #739, opened 2022-05-17
Improve Ender5Plus config

alex255:
### Description

upgraded and tested config for ender 5 plus and enable the new and usefull features of the Marlin 2

### Benefits

- set max hotend temp to 300 because the base hotend can handle this without problem.
- enable and config PIDTEMPBED
- increase max extrude lenght for longer bowden and filament change
- modify Z offset to correct value
- slow down the leveling and home Z speed ( the default is too fast for the machine )
- enable Z_MIN_PROBE_REPEATABILITY_TEST
- enable FILAMENT_RUNOUT_SCRIPT
- always enable auto leveling after home
- enable PREHEAT_BEFORE_LEVELING
- enable HOTEND_IDLE_TIMEOUT
- modify unload/load lenght to the real bowden lenght
- modify ADVANCED_PAUSE_FEATURE settings, and now work perfectly with and without screen (octoprint) 
- enable more M114 report options for octoprint


unitrix0:
Did you test that BLTOUCH_HS_MODE still works? If I set the probe speed lower thant 10*60 it fails.

alex255:
yes, it worked for me without problem, but for safety i disabled it

config/examples/Creality/Ender-5 Plus/Configuration.h
```diff
@@ -612,14 +612,14 @@
...
-#define HEATER_6_MAXTEMP 275
-#define HEATER_7_MAXTEMP 275
+#define HEATER_0_MAXTEMP 300
+#define HEATER_1_MAXTEMP 300
+#define HEATER_2_MAXTEMP 300
+#define HEATER_3_MAXTEMP 300
+#define HEATER_4_MAXTEMP 300
+#define HEATER_5_MAXTEMP 300
+#define HEATER_6_MAXTEMP 300
+#define HEATER_7_MAXTEMP 300
```
thinkyhead: Does this come with an all-metal hotend as a standard component? It looks like it does. If it doesn't, then the max temp should be set below the off-gassing temperature of PTFE, because the fumes are dangerous.

thinkyhead: Creality shows 260°C as the max allowed hotend temp, which is still a little high, but safer when PTFE is involved.

PR #747, opened 2022-06-05
Enable PID Autotune menu, Sound on/off menu, Remaining percent on LCD

0Hertz:
### Requirements

### Description

Enable PID Autotune menu, Sound on/off menu, Remaining percent on LCD.

### Benefits

You can enable disable sound from LCD, autotune PID  directly from LCD and view remaining time from info screen.


ellensp:
These are personal preferences, why should they be applied to everyone? 
Are they enabled on the original machine?

0Hertz:
On the original firmware you can see the percentage, on my printer with the original firmware i had problem with the hotend temperature stabilization (plus minus 3 degrees) and i had to do PID calibration with my PC, it is easier to do it from menu,
and for the sound ON/OFF i believe it is something that it not making a problem if it is there.

thinkyhead:
I am okay with this contribution. I generally think we should use all the available space on the board for helpful features that improve the user experience, and not necessarily leave things out that were not in the OEM configuration. Otherwise it's just wasted unused flash space. For example, the tramming wizard and probe Z offset wizard are helpful and should be enabled when there is available space. I think that providing user-enhanced and user-tuned configurations helps to make users see the open source Marlin project as a real upgrade and even a fix compared to the more limited firmware they get from the manufacturer.

PR #743, opened 2022-05-29
Fix Ender-6 BLTouch Option

thisiskeithb:
### Description

Fix Ender-6 BLTouch Option

### Benefits

BLTouch option will work correctly.

### Related Issues

https://github.com/MarlinFirmware/Configurations/pull/737#issuecomment-1140423514

PR #749, opened 2022-06-07
Fix Ender-6 BLTouch Option

thisiskeithb:
### Description

Redoing #743 since it was trashed through some force-pushes of the import-2.1.x branch.

Fix Ender-6 BLTouch Option

### Benefits

BLTouch option will work.

### Related Issues

#743


thisiskeithb:
This should also apply to 2.0.9.4 and 2.1, so let me know if you need PRs for those.

PR #752, opened 2022-06-16
Add Teensy35_36_SIMPLE on LulzBot Mini2_Hibiscus and Mini2_Hibiscus_Experimental

xxxajk:
See Configuration.h for details.

### Requirements
As listed in Configuration.h:

```
/**
 * NOTE! This config file is a bare-bones starting point for replacing
 * the RAMBO board in the LULZBOT Mini2 (Hibiscus_Experimental)
 *
 * NOTE! The hardware for this particular config is for a LULZBOT Mini 2 Hibiscus,
 * with an CecropiaSilk_SingleExtruderAeroV2 and the TEENSY35_36_SIMPLE layout.
 * *ANY* board/cpu can be used with the same config file
 *        *providing the following conditions are met*
 * The original 24V power supply, SingleExtruderAeroV2, 2.85mm as the toolhead and
 * the Z Axis is using a belt and belt-brake remain, and that you use TMC2130 stepper drivers.
 * The same config has been used with an AVR 2560 and RAMPS 6+.
 *
 * NOTE!
 * You *WILL LOSE THE LCD DISPLAY*, this is intentional, so that you can add
 * whatever LCD you want to.
 * The largest difference overall is the resolution for this printer has been doubled.
 * The reason is because I like to sometimes print using 0.2 and 0.1mm nozzles.
 */

```
### Description
See above.

### Benefits
Allows replacement of AVR with something better.

### Related Issues
None.


thisiskeithb:
Did you start with a fresh / default config from the `import-2.1.x` branch? You're missing these lines for example:

https://github.com/MarlinFirmware/Configurations/blob/c27da0bf6a3ff090851a350cc2a897668b4dc85c/config/default/Configuration.h#L23-L24

Also, Configs are also sorted by OEM -> Model, so these should be moved under `/config/examples/LulzBot` folder.

This also needs https://github.com/MarlinFirmware/Marlin/pull/24318 to be merged first.

xxxajk:
Yes, it's from a fresh pull, but the config isn't, I'll fix it

xxxajk:
This config also directly fits into 2.1-bugfix

thisiskeithb:
Bugfix strips out those errors automatically. 

They need to be there in the import branches.

Please fix it.

xxxajk:
:-) Didn't know that. Nice! and Done!

config/examples/Teensy/35_36_SIMPLE/Mini2_Hibiscus_Experimental/Configuration.h
```diff
@@ -0,0 +1,3114 @@
...
+
+// Enable the Bluetooth serial interface on AT90USB devices
+//#define BLUETOOTH
+
+// Name displayed in the LCD "Ready" message and Info menu
+#define CUSTOM_MACHINE_NAME "LulzBot Mini 2"
+
+// Printer's unique ID, used by some programs to differentiate between machines.
+// Choose your own or use a service like https://www.uuidgenerator.net/version4
+#define MACHINE_UUID "e5502411-d46d-421d-ba3a-a20126d7930f"
```
thisiskeithb: This should be reset/disabled since it's unique to you/your printer.

thisiskeithb:
It is important that you start with a fresh/clean copy of `import-2.1.x` because you're missing these from `Configuration.h`:

https://github.com/MarlinFirmware/Configurations/blob/c27da0bf6a3ff090851a350cc2a897668b4dc85c/config/default/Configuration.h#L2733-L2736

and

https://github.com/MarlinFirmware/Configurations/blob/c27da0bf6a3ff090851a350cc2a897668b4dc85c/config/default/Configuration.h#L2763-L2767

You're also missing these from `Configuration_adv.h`:

https://github.com/MarlinFirmware/Configurations/blob/c27da0bf6a3ff090851a350cc2a897668b4dc85c/config/default/Configuration_adv.h#L1665-L1672

and

https://github.com/MarlinFirmware/Configurations/blob/c27da0bf6a3ff090851a350cc2a897668b4dc85c/config/default/Configuration_adv.h#L3891-L3897

and

https://github.com/MarlinFirmware/Configurations/blob/c27da0bf6a3ff090851a350cc2a897668b4dc85c/config/default/Configuration_adv.h#L4442-L4443

config/examples/Teensy/35_36_SIMPLE/Mini2_Hibiscus_Experimental/Configuration.h
xxxajk: Actually, it's not unique for the brand... every one of this particular printer has the same UUID, so that the printing software can sort out the features, and that's why it is set. It's actually needed to use on patched lulzbot cura software, or features are disabled in the UI.

xxxajk:
I'll have to hand patch those manually, not a big deal. I'll push those changes shortly. Thanks for finding these. 
I actually started with the 2.1-bugfix configs... So I don't know why they would be out of date, unless the zip file was behind.

config/examples/Teensy/35_36_SIMPLE/Mini2_Hibiscus_Experimental/Configuration.h
xxxajk: Also, worse, they add another UUID to id the tool head, which I did not add, because I don't use the stock nozzle. Their software happily ignores it if missing, which is a good thing, just pukes a harmless warning in the log, and continues.
if the UUID isn't correct, it will refuse to use the machine.... it's bad enough I had to hack in the USB ID for it to be recognized...

cp2004:
> It is important that you start with a fresh/clean copy of import-2.1.x

> I actually started with the 2.1-bugfix configs... 

The [import-2.1.x](https://github.com/MarlinFirmware/Configurations/tree/import-2.1.x) configs are different to bugfix-2.1.x

xxxajk:
Right, Not a big deal, they're patched, updating them...

xxxajk:
Let me know if I missed/messed up anything else, but the UUID is required, which is why it's in a deeply nested directory specifically naming the machine.

thisiskeithb:
> The [import-2.1.x](https://github.com/MarlinFirmware/Configurations/tree/import-2.1.x) configs are different to bugfix-2.1.x

@cp2004 That is correct and intentional. 

Through the magic of scripting/deployment, pending features & “Do not build with…” lines are removed when changes are deployed to bugfix.

thisiskeithb:
> but the UUID is required, which is why it's in a deeply nested directory specifically naming the machine

This still needs to be moved to the main Lulzbot folder.

xxxajk:
Is there one? Problem is that it's no longer really a lulzbot machine... it's Teensy based now, just uses the lulzbot frame, and none of the electronics except for the wiring harness, and steppers/sensors.

xxxajk:
Oh yeah, and worse, it was a prototype, never released! Notice that I disabled any use of the display too, so a regular user is going to be very confused as to why it no longer displays anything... the driver for the display must be heavily patched, because of the software spi implementation trying to move data at 30MHz on this particular board... and I'm not going to be using that anyway.

thisiskeithb:
> Is there one? 

Yes: https://github.com/MarlinFirmware/Configurations/tree/import-2.1.x/config/examples/LulzBot/

> Problem is that it's no longer really a lulzbot machine...

See the Ender-3 config folder (or any other Creality config) where it has the most configs/upgrades: https://github.com/MarlinFirmware/Configurations/tree/import-2.1.x/config/examples/Creality/Ender-3

The standard patten is OEM -> Model -> Boards/Upgrades

It would also help to add a `README.md` in the same folder as your new configs explaining the differences over a stock machine so we aren't stuck trying to troubleshoot it if/when something doesn't work down the road.

thisiskeithb:
> Oh yeah, and worse, it was a prototype, never released! 

So should this config really be added then or is this just a one-off for your personal use?

xxxajk:
Good idea, I'll add the readme, I'm sure others out there who want to upgrade easily DIY. 
Currently I have the only machine with the hardware replaced, however, there's no real examples at all for Teensy, and that ends up making this modification difficult for the average user. they've got no place to start! TAZ-4 != the Mini2.
I'm sure many others have tried to replace the board, only to be disappointed, and this opens up the avenue for the replacement. I'll also at some point have a board offering that allows this modification anyway, posted on github...

So... the TLDR is:
Basically this isn't an "approved modification" and isn't anything that will ever be commercially supplied. The config *does* give others a starting point, and increases the resolution on the X/Y/Z axis over the stock old firmware in any case. there isn't one single Teensy *anything* example, so my goal here is to give a starting point for others.

thisiskeithb:
> TAZ-4 != the Mini2.

I never said it was. I am saying that you need to move your config to the Lulzbot folder.

xxxajk:
That's fine, but won't it cause confusion? Are there any other taz-4's that have the guts torn out and replaced? 
I'm guessing not... It's already 32bit :-) As long as you think it won't cause owners of the Mini2 any confusion as to why it doesn't work on the stock RAMBO or einsy retro board, I'm fine with the move... and No, I can't do the mods for that board, since the whole reason I'm replacing it is that the AVR is too slow for double resolution, and the RAMBO board fried, and I can't justify the cost of replacment, especially since I can build my own hardware...

xxxajk:
Hopefully this won't cause any confusion, I'll put a README.md for Teensy (and name it Teensy boards) so that people can at least locate an example for the board if you think that would be a good idea.

config/examples/Teensy/35_36_SIMPLE/Mini2_Hibiscus_Experimental/Configuration.h
InsanityAutomation: This was done to identify when firmware needed to be flashed for tool head specific slicer profiles. Something some of the larger corporate customers wanted I would assume. I concur that it should stay.

thinkyhead:
@xxxajk — This one still doesn't build because there's no `BOARD_TEENSY35_36_SIMPLE` existing. Where can we acquire the pins definition for this board, or should we use one of the existing board definitions?

PR #757, opened 2022-06-19
Fix Biqu BX Config

thisiskeithb:
### Description

Fix Biqu BX serial ports because:

- Serial port 1 is for the TFT
- Serial port -1 is for the micro usb
- Serial port 4 is for the WiFi header

Also, disable `TOUCH_SCREEN`.

Found while working with @rhapsodyv on https://github.com/MarlinFirmware/Marlin/issues/24353

We'll still need a PR to fix the BX booting issue (hopefully coming soon), but this is also required.

### Benefits

Correct serial ports.

### Related Issues

- https://github.com/MarlinFirmware/Marlin/issues/24353

### DIFF
--- a/config/examples/BIQU/BX/Configuration.h
+++ b/config/examples/BIQU/BX/Configuration.h
@@ -2946,7 +2946,7 @@
 //
 // Touch Screen Settings
 //
-#define TOUCH_SCREEN
+//#define TOUCH_SCREEN
 #if ENABLED(TOUCH_SCREEN)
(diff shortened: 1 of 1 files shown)


thisiskeithb:
Closing until an upstream PR to fix BX issues is submitted.

thisiskeithb:
Ok. Serial ports have been confirmed as noted above & this config should be updated.

thisiskeithb:
The last commit disables TOUCH_SCREEN support since it causes bootloops/lockups.

PR #763, opened 2022-06-22
More Biqu BX Improvements

thisiskeithb:
### Description

- Marlin will silently set `ENCODER_PULSES_PER_STEP` to 4 if not defined, but this was too coarse on mine & rhapsodyv's setups, so it was changed to 5.
- Explicitly set `SDCARD_CONNECTION` to `ONBOARD`. Marlin already does this silently, but now it's clear in the config. I would not recommend using the USB or SD card slots on the TFT due to possible data corruption.
- Individual Axis Homing
- Cleanup / improve the `BABBYSTEPPING` section since there was a duplicate line.

### Benefits

Menu items won't be skipped/finer adjustment via encoder will work, clearer config.

### Related Issues

https://github.com/MarlinFirmware/Marlin/issues/24353

### DIFF
--- a/config/examples/BIQU/BX/Configuration.h
+++ b/config/examples/BIQU/BX/Configuration.h
@@ -2273,7 +2273,7 @@
 // This option overrides the default number of encoder pulses needed to
 // produce one step. Should be increased for high-resolution encoders.
 //
-//#define ENCODER_PULSES_PER_STEP 4
+#define ENCODER_PULSES_PER_STEP 5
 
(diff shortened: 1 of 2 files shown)


thisiskeithb:
The last commit explicitly defines `SDCARD_CONNECTION` as `ONBOARD` to make it clear that the onboard SD card slot is in use. I would not recommend using the USB or SD card slots on the TFT due to possible data corruption.

PR #768, opened 2022-06-27
Configuration files for Velleman Vertex Delta K8800

sgparry:
### Description

This PR provides configuration files that allow Bugfix_2.1.x of Marlin to be built for the Velleman Vertex Delta K8800 printer. It provides two variants -one with Bilinear Levelling and one with Unified Bed Levelling

### Benefits

The stock firmware for the K8800 is based on 1.1.4, which is not as reliable as 2.x. The stock firmware is commonly blamed for frequent bed to head crashes and belt slippage and uses bilinear levelling rather than UBL. The stock firmware also includes a number of hacks to make it work with the hardware and is (not) maintained separately from Marlin. No changes have been made in two years.

These updated configurations yield a firmware that works with unaltered core Marlin 2.1.x - they are more robust and more easily maintainable and appear to dramatically reduce the head crashes and belt slippage.

### Related Issues

https://github.com/MarlinFirmware/Configurations/issues/649#issuecomment-1166694610


thisiskeithb:
It looks like you started with a `release-2.1.x` config since you’re missing the standard “Don't build with import-2.1.x configurations!” error lines at the top of the configs and have `CONFIG_EXAMPLES_DIR` lines (these are added automatically through the magic of scripting in `bugfix-*` & `release-*` branches) as well as some other missing blocks of code/config options found in `import-2.1.x`.

I’d recommend starting over with the [default delta config](https://github.com/MarlinFirmware/Configurations/tree/import-2.1.x/config/examples/delta/generic) and adding your config options so they comply with the standard `import-2.1.x` config format.

sgparry:
Thanks for looking at this @thisiskeithb - sorry, I was not familiar with the drill.
The problem I have now is that this configuration includes literally hundreds of lines of changes, collated from four different sources. I think I will try to generate a patch of the changes I have made, apply it to the default delta config, review the results line by line and then test.
Can I just commit the resulting configuration to this PR or do I have to raise a new one please?

thisiskeithb:
> Can I just commit the resulting configuration to this PR

That works.

thinkyhead:
Unable to push any changes to your fork, possibly because it is called "Configurations2022". So I will close this PR and push the changes directly.

sgparry:
Hi @thinkyhead 
Thanks for recovering and merging this PR. I noticed two issues though:
- Only the Bilinear variant made it into the import-2.1.x branch, you left out the UBL variant - should I raise a new PR for the UBL configuration?
- The commit for my stuff includes some changes for the foam cutter.... not sure if that was intentional?

thisiskeithb:
> you left out the UBL variant - should I raise a new PR for the UBL configuration?

You'll need to submit another PR.

> The commit for my stuff includes some changes for the foam cutter.... not sure if that was intentional?

That tends to happen around here.

sgparry:
@thisiskeithb @thinkyhead - I thought the UBL configuration was left out, but actually TH merged the configurations and included a define to select between them. There are two resulting issues:
- the `#define` default is currently UBL, whereas the folder is called K8800-Bilinear; Bilinear would be the safer default.
- the UUID for the variants is the same - I was deliberately using different UUIDs to allow a slicer / host to quickly differentiate between the two variants - not sure if the back-end scripts will like this.

Which would you prefer me to do in a new PR:
A: Keep the single configuration, correct the default to BiIinear but change the UUID in an `#if`.
-or-
B: Create a separate configuration for UBL, with its own UUID.

Issue #771, opened 2022-07-03
[BUG] K8800 bed levelling default inconsistent and hazardous.


sgparry:
### Bug Description

The newly added configuration for the K8800 has a default bed levelling that is inconsistent, confusing and possibly hazardous.
The original Pull Request included two configurations - one for bilinear leveling, one for UBL. On commit, these were merged together using `#define` and `#if`, but this introduced some problems:

- the `#define` default is currently UBL, whereas the folder is called `K8800-Bilinear`; Bilinear would be the safer default (UBL requires the printer definition in the slicer to be changed; failure to do so can cause bed-to-head crashes.).
- The UUID for the variants is the same - Different UUIDs were initially defined to allow a slicer / host to quickly differentiate between the two variants.
- Anyone wanting to use the non-default variant will have to build it themselves; many users rely on automated build services to update their Marlin firmware.
It did however have the advantage of only adding one profile to Marlin rather than two.

### Configuration Files
[Configuration.zip](https://github.com/MarlinFirmware/Configurations/files/9035230/Configuration.zip)

### Steps to Reproduce

- Build 2.1.1 firmware using the K8800-Bilinear configuration
- Upload to your K8800
- Slice a test model with Cura using default settings for K8800
- Print the model.
or
- look in Configuration.h at around line 31 and compare `#define K8800_*` settings with the name of the folder
- look in Configuration.h at around line 154 at the UUID `#define`.

**Expected behavior:**

Default configuration should build as bilinear levelling, or as both bilinear and UBL. The bilinear variant should work as a drop in replacement for the stock 1.1.4 firmware.

**Actual behavior:**

The print will either stall with an error or head crash just after the purging part of preflight.

#### Additional Information

I will submit two mutually exclusive PRs to fix this - one for a combined configuration, one for split. The project owners are free to adopt whichever one (or none) they please.

thinkyhead:
It is much preferred to have fewer configurations, not a different configuration for every potential variant, whether that consists of a different probe, a different leveling system, or a different [your peripheral here]. If a different UUID is needed for the two types of bed leveling, and this is the primary means of distinguishing these variants, then the UUID may be applied like the other settings that are contingent on the bed leveling system. If there are other settings that need to differ based on the form of bed leveling, those may also be applied in your next PR.

Unfortunately your PRs are not editable by admins. This could be due to a glitch in GitHub, but double-check the "allow edits by maintainers" checkbox in your next submission.

[...]

PR #772, opened 2022-07-03
Possble Fix B for #771 [BUG] K8800 bed levelling default inconsistent and hazardous. 

sgparry:
### Description

This is one of two possible fixes (A and B) to bug #771; They are mutually exclusive - please only merge one of them.
This sets up separate configurations for Bilinear and UBL configurations. This corrects the confusion reported in issue #771 and keeps the UUIDs distinct for the two variants

### Benefits

Fixes the original issue; also allows the two configurations to be built side by side. It also fixes compiler issues with the boot screen acquired during the merge; also makes the configuration cosmetically closer to the examples, reducing the 'noise' that appears when differencing.

### Related Issues

Issue #771


GastonMehaudy:
Hi, I'm from Argentina, I have a tenlog td3 pro printer. I installed the luxuri firmware 2.1 it works fine but for other reasons I want to go to 32bit but I can't buy abroad due to problems in my country. I got a bigtreetech octopus v1.1 with a tft70 screen. can you do that update? Please, if you can confirm the purchase, since it is the only thing I get in 32bit and that it has for various engines.

sjasonsmith:
I am concluding the following based on this sitting almost two years with no activity:

1. The better of these two solutions was not clear or correct enough to merge quickly.
2. This is not a major issue since nobody else has commented on either PR or the referenced issue.

I am going to go ahead and close this without merging.

PR #773, opened 2022-07-03
Possble Fix A for #771 [BUG] K8800 bed levelling default inconsistent and hazardous. 

sgparry:
### Description

This is one of two possible fixes (A and B) to bug #771; They are mutually exclusive - please only merge one of them.
This fixes the combined configuration containing Bilinear and UBL options. This corrects the confusion reported in issue #771 and keeps the UUIDs distinct for the two variants

### Benefits

Fixes the original issue; keeps everything in one configuration. It also fixes compiler issues with the boot screen acquired during the merge; also makes the configuration cosmetically closer to the examples, reducing the 'noise' that appears when differencing.

### Related Issues

Issue #771


sjasonsmith:
I am concluding the following based on this sitting almost two years with no activity:

1. The better of these two solutions was not clear or correct enough to merge quickly.
2. This is not a major issue since nobody else has commented on either PR or the referenced issue.

I am going to go ahead and close this without merging.

PR #776, opened 2022-07-06
Put DELTA / SCARA / TPARA in all configs

thinkyhead:
To make all configurations identical in form, include DELTA, SCARA, and TPARA options in all configurations.


config/default/Configuration.h
```diff
@@ -867,6 +858,166 @@
...
+
+  #define SCARA_FEEDRATE_SCALING  // Convert XY feedrate from mm/s to degrees/s on the fly
+
+  // Radius around the center where the arm cannot reach
+  #define MIDDLE_DEAD_ZONE_R   0  // (mm)
+
+  // Calculated from Calibration Guide and M360 / M114. See http://reprap.harleystudio.co.za/?page_id=1073
+  #define THETA_HOMING_OFFSET  0
+  #define PSI_HOMING_OFFSET    0
+#endif
```
sjasonsmith, line 987: The AXEL_TPARA section is duplicated. I've only looked at this one file but I assume it is the same elsewhere.

config/default/Configuration.h
```diff
@@ -1241,9 +1392,38 @@
...
-//
-// For Z_PROBE_ALLEN_KEY see the Delta example configurations.
-//
+/**
+ * Allen key retractable z-probe as seen on many Kossel delta printers - https://reprap.org/wiki/Kossel#Automatic_bed_leveling_probe
+ * Deploys by touching z-axis belt. Retracts by pushing the probe down.
+ */
+//#define Z_PROBE_ALLEN_KEY
+
+#if ENABLED(Z_PROBE_ALLEN_KEY)
```
sjasonsmith, line 1368: This probably belonged in this file anyway... even if it was _commonly_ used by delta printers, there's no reason it couldn't be used by other types with the right config.

sjasonsmith:
I just looked at the default configuration file, but I think this would be a positive change. It allows the same processes to be used for all printers, without extra steps needed unique to non-cartesian printers.

Once the duplicate section I mentioned is fixed, it looks like this is only growing the file by 153 lines. Given that the config file was already ~1400 lines I don't see that as a problem. The scale on config confusion probably won't increase, since these sections are easily ignored by people not using them.

sjasonsmith:
It looks like for now you are just pushing the added sections out to all the non-delta configs?

I imagine re-working the delta configs themselves is going to be more hands-on, especially where multiple variants exist in one file (such as for the AnyCubic Kossel LP)

config/default/Configuration.h
thisiskeithb: I was working with a vendor last year that used this probe type on a cartesian machine, so they will probably be happy that they no longer have to port it over/out of delta configs each update.

thinkyhead:
Ah, I see you have been commenting. Things are now in good shape after cleanup, in spite of the initial expected search/replace imperfections. Also fixed up a number of lingering errors.

PR #794, opened 2022-08-07
Create FT-5 R2

MasterofNull:
### Requirements

Adding configuration files for Folger Tech FT-5 R2. With bed size and stop locations for the original configuration.


thisiskeithb:
There's nothing in this PR other than a file with a single blank line in it: https://github.com/MarlinFirmware/Configurations/pull/794/files

Issue #805, opened 2022-08-23
Too many icons for Creality Ender Display board rev1.2


Ergonomicmike:
### Bug Description

When - to use the Jyers' UI - I flashed the 'private' file to my Ender 3 V2 Display with a rev 1.2 PCB, the flash fails about 7/8th's into the flash operation.

### Steps to Reproduce

If you have a Display with a rev 1.2 PCB, try it yourself. Copy 'private' from
Configurations/config/examples/Creality/Ender-3 v2/LCD files 
into your SD card, put your card in your Display, power up and wait.

You'll see the icon files being saved one by one, Until it stops.

**Expected behavior:**

Naturally all the icons should be saved and the flash operation should complete successfully.

**Actual behavior:**

The flash operation fails with an "Error."

#### Additional Information

I suspect that Creality 'cheaped out' on memory on this version of the Display board. 

Because I was able to load the icons that I needed by deleting the icon set with a 3 in it, thus reducing the memory footprint.

I also notice that every icon comes in a pair. I don't know if this was by design or a mistake. If the latter, then the obvious solution is to eliminate the duplicate icon, reducing the memory footprint.

ellensp:
There are now 3 different types of this display. (that we have seen thus far) https://marlin.crc.id.au/forum/t/identifying-ender-3-v2-screen-versions/960
the 2nd and 3rd are completely undocumented

ellensp:
"icon comes in a pair."

All icons are unique 

Some are sutile having a darker or lighter background colour for when icon is selected or not.

Ergonomicmike:
Attached is a photo of my Rev 1.2 Board, to add to your collection.

It has the F1C100s also. And it also requires "private" to upload files to it.
![Creality Rev 1 2 Display Board](https://user-images.githubusercontent.com/15025888/186068133-964ec722-f80c-443f-99e8-2c4e5f4ba179.jpg)


On the "pair" of icons: Okay. that makes sense. (It wasn't obvious to me on my monitor.) 

Thanks for the quick reply.

ellensp:
It is unlikely to be the number of icons, more likely to be the total space they take up

try replacing the icons with these, they have been a bit better packed.

[repacked-icons.zip](https://github.com/MarlinFirmware/Configurations/files/9399648/repacked-icons.zip)

also try replacing files in image 

[repacked-images.zip](https://github.com/MarlinFirmware/Configurations/files/9399702/repacked-images.zip)

[...]

PR #807, opened 2022-08-23
repack private jpegs with jpeg.io

ellensp:
### Description

Image data in private data directory is two large to fit into DACAI  e3v2 lcd clone (F1C100s mpu)

Repacked all jpeg data with jpeg.io online tool. They now fit.

### Benefits

Fit into DACAI e3v2 lcd clone flash memmory.

### Related Issues
https://github.com/MarlinFirmware/Configurations/issues/805


thinkyhead:
Some of these might look better if they were compressed from an original clean image. I can provide those.

ellensp:
I could not see any visible difference...

Sparticuz:
@ellensp I'm assuming this is in release-2.1.2, but I received this trying to flash.
image

Sparticuz:
It looks like because I used an iPad and it created a ton of . (dot) files.

PR #812, opened 2022-09-17
Insert Nano V1.3 board to select in BlueR Gcode

ils15:
Some BlueR arrived with MKS Robin Nano V 1.3

This adjust alow to user quick select thus board by uncomment file

### DIFF
--- a/config/examples/Two Trees/BlueR/BlueR V3/Configuration.h
+++ b/config/examples/Two Trees/BlueR/BlueR V3/Configuration.h
@@ -20,11 +20,10 @@
  *
  */
 #pragma once
-#error "Don't build with import-2.1.x configurations!"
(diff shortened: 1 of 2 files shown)


config/examples/Two Trees/BlueR/BlueR V3/Configuration.h
```diff
@@ -20,11 +20,10 @@
  *
  */
 #pragma once
-#error "Don't build with import-2.1.x configurations!"
-#error "Use the 'bugfix...' or 'release...' configurations matching your Marlin version."
```
thisiskeithb: These lines should not be removed. Please revert this change.

Joeben349:
Cashapp login issues

PR #819, opened 2022-09-28
VORON V2 => V1

thisiskeithb:
### Description

The [VORON V1 (Legacy)](https://vorondesign.com/voron_legacy) was originally ported from a Marlin 1.1.8 config and added as the V2, which was incorrect.

### Requirements

VORON Legacy (V1, RAMPS-based)

### Benefits

Correct VORON Version/Name

### Related Issues

- https://github.com/MarlinFirmware/Marlin/pull/24811
- https://github.com/MarlinFirmware/Marlin/pull/12705#issuecomment-452720428


thisiskeithb:
Apparently there is a difference between the VORON 2.0 (ramps, much older / based off Marlin 1.1.0 RC8) and the 2.4 which is the higher-end / current CoreXY.

See https://github.com/VoronDesign/Voron-2/tree/Voron2.0/Firmware/Marlin

PR #831, opened 2022-10-20
TronxyX5S-2e with Chitu cxy-v5-180409

silycr:
Standard configuration
Included support for upgrades to WIFI(ESP8266) and BLTOUCH / Inductive

It appears the Tronxy X5S-2e was availiable in 2 configurations, with an 8-bit controller/red MB in early models . This configuration supports the STM32f1 32-bit board, with cxy-v5-180409 written on the green MB.

### Requirements

Nil

### Benefits

Added model for support

### Related Issues

Nil


silycr:
https://github.com/MarlinFirmware/Marlin/pull/24910
Pull request for BLTouch pin assignment

PR #833, opened 2022-10-21
Tevo Tornado needs ENDSTOP_NOISE_THRESHOLD

shorage:
line 1142 uncomment, it is NEEDED
#define ENDSTOP_NOISE_THRESHOLD 2 //Needed for stock Tevo Tornado

Description

line 1142 uncomment, it is NEEDED
#define ENDSTOP_NOISE_THRESHOLD 2 //Needed for stock Tevo Tornado
Requirements

Benefits

I can home without printer fault

Related Issues

Speradic printer faults during auto home

### DIFF
--- a/config/examples/Tevo/Tornado/V1 (MKS Base)/Configuration.h
+++ b/config/examples/Tevo/Tornado/V1 (MKS Base)/Configuration.h
@@ -1139,7 +1139,7 @@
  *
  * :[2,3,4,5,6,7]
  */
-//#define ENDSTOP_NOISE_THRESHOLD 2
+#define ENDSTOP_NOISE_THRESHOLD 2
 
 // Check for stuck or disconnected endstops during homing moves.
 //#define DETECT_BROKEN_ENDSTOP
--- a/config/examples/Tevo/Tornado/V2 (MKS GEN-L)/Configuration.h
+++ b/config/examples/Tevo/Tornado/V2 (MKS GEN-L)/Configuration.h
@@ -1139,7 +1139,7 @@
  *
  * :[2,3,4,5,6,7]
  */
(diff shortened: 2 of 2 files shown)


thisiskeithb:
The V1 / MKS Base variant should be updated as well since `ENDSTOP_NOISE_THRESHOLD` is disabled:

https://github.com/MarlinFirmware/Configurations/blob/c5806c0a300a794f7a934c26f781d23643a423d1/config/examples/Tevo/Tornado/V1%20(MKS%20Base)/Configuration.h#L1142

thinkyhead:
> The V1 / MKS Base variant should be updated as well since `ENDSTOP_NOISE_THRESHOLD` is disabled…

Most important if it's an endstop issue and not a board issue, but it never hurts to enable this on any machine.

thisiskeithb:
> Most important if it's an endstop issue and not a board issue,

I never ran into it on my original Tornado with an MKS Base or when I upgraded it to a Gen-L, but like you said, it doesn't hurt to enable it.

shorage:
> > Most important if it's an endstop issue and not a board issue,
> 
> I never ran into it on my original Tornado with an MKS Base or when I upgraded it to a Gen-L, but like you said, it doesn't hurt to enable it.

Fair enough, i did test my endstops and they checked good. It prints like a beast with this enabled. Without it i cant get passed homing. Took me 2 days and reviewed the config hundreds of times before i enabled it. If it saves someone some time, i think it is worth it. It would randomly choose to fault without it enabled during the homing process or x-y movement and it would never be in a consistent spot. Tested a bunch of end-stops i had laying around. I have Marlin 1.1.9 for the longest and i noticed this was enabled. Then tried it. Now all is well.

Regards,

Frank

PR #837, opened 2022-11-03
Fix Biqu B1 SE Plus Fan Config

thisiskeithb:
### Description

Based on https://github.com/MarlinFirmware/Marlin/pull/24995, enable both Controller fans and limit number of `M106` fans to only those that are connected.

For reference:

`FAN1_PIN` = Controller fan on side of case
`FAN2_PIN` = Controller fan near drivers

### Benefits

Better user experience & automated controller cooling since it was only enabled manually with the current config.

### DIFF
--- a/config/examples/BIQU/B1 SE Plus/Configuration.h
+++ b/config/examples/BIQU/B1 SE Plus/Configuration.h
@@ -3238,7 +3238,7 @@
 
 // Set number of user-controlled fans. Disable to use all board-defined fans.
 // :[1,2,3,4,5,6,7,8]
-//#define NUM_M106_FANS 1
+#define NUM_M106_FANS 1
 
 // Use software PWM to drive the fan, as for the heaters. This uses a very low frequency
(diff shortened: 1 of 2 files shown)


thisiskeithb:
I dug around [Biqu’s SE Plus source code](https://github.com/bigtreetech/BIQU-B1-SE-PLUS/blob/master/B1-SE-PLUS%20firmware/B1-SE-SKR2(2.6.0).rar) and discovered that they hacked in dual controller fan support, so this PR will only activate the fan on the outside of the case (FAN1) and not the additional fan (FAN2) without mods upstream in Marlin.

Their mod for dual fan support was pretty quick and dirty, so it’ll take some additional changes to make it “merge ready” upstream, but at least this PR will get some fresh air inside the case.

thisiskeithb:
Having https://github.com/MarlinFirmware/Marlin/pull/24995 merged would allow both controller fans to work, so I’d need to push an additional config change if/when that occurs before this is merged.

PR #853, opened 2022-12-07
Add Skr3 ender5 plus example config from rubienr

ellensp:
### Requirements

Ender 5 plus with stock touch LCD, A4988 stepper drivers

### Description

Example Config for this machine and controller.


ellensp:
github doesn't like that config/examples/Creality/Ender-5 Plus/Configuration.h was updated after this PR moved it to a sub-directory...

thinkyhead:
> github doesn't like that config/examples/Creality/Ender-5 Plus/Configuration.h was updated after this PR moved it to a sub-directory...

A `git merge` should recognize the rename and do the right thing. I just did that to update this PR and it didn't give any complaints.

PR #856, opened 2022-12-08
Ender 3 v2 Neo Configs

ellensp:
### Requirements

Ender 3 v2 Neo 

### Description

Add Ender 3 v2 Neo Configs

### Benefits

Have example config for Ender 3 v2 Neo

### DIFF
--- a/config/examples/Creality/Ender-3 V2 Neo/Configuration.h
+++ b/config/examples/Creality/Ender-3 V2 Neo/Configuration.h
@@ -1,6 +1,6 @@
 /**
  * Marlin 3D Printer Firmware
- * Copyright (c) 2020 MarlinFirmware [https://github.com/MarlinFirmware/Marlin]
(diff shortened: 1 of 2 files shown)


benbenwilde:
Couple things is that with this config it doesn't auto home currently when you try to move xyz, which is what the stock firmware did on v2 Neo. obviously unsafe since you can tell it to go past boundaries if it was moved manually. Same for adjusting z-offset, which needs some baby stepping settings to work, but possibly not all of the ones that I turned on. One or two of them might make it less safe as well. The stock firmware also triggered an auto home when trying to change z-index. Seems like there would be a setting for this somewhere.

The-EG:
@benbenwilde There are `NO_MOTION_BEFORE_HOMING` and `HOME_AFTER_DEACTIVATE` that can be enabled, although I would say it's a personal preference. Having both disabled by default seems to be convention since several other Ender 3 examples are that way as I would expect many other examples here.

Users that want that behavior can enable those and other options at will.

PR #872, opened 2022-12-31
Update ASSISTED_TRAMMING section of Configuration_adv.h for Ender 3 Pro

zanhecht:
### Requirements

### Description

On Ender 3 Pro Configuration_adv.h, wrap ASSISTED_TRAMMING_WIZARD in a check for UI, and set bed dimensions and default screw thread to correct values for Ender 3 Pro.

### Benefits

Makes sure the ASSISTED_TRAMMING_WIZARD cannot be enabled on printers without a UI.

Provides correct values for screw size and probe positions for this specific printer model if the option is enabled.


thisiskeithb:
> Makes sure the ASSISTED_TRAMMING cannot be enabled on printers without a probe.

[Marlin has a sanity check for this already, so there's no need to do that](https://github.com/MarlinFirmware/Marlin/blob/4b4d98d282f2760e760b9f2ccca8dcc8831f7f8d/Marlin/src/inc/SanityCheck.h#L2856-L2861):

```cpp
/**
 * G35 Assisted Tramming
 */
#if ENABLED(ASSISTED_TRAMMING) && !HAS_BED_PROBE
  #error "ASSISTED_TRAMMING requires a bed probe."
#endif
```

> Provides correct values for screw size and probe positions for this specific printer model if the option is enabled.

The screw holes on the Ender-3/5 beds are inset by 32.5 mm from each side, so you might as well update that if you want to prefill this section:

```diff
- #define TRAMMING_POINT_XY { {  20, 20 }, { 180,  20 }, { 180, 180 }, { 20, 180 } }
+ #define TRAMMING_POINT_XY { {  32.5, 32.5 }, { 202.5,  32.5 }, { 202.5, 202.5 }, { 32.5, 202.5 } }
```

zanhecht:
@thisiskeithb I removed the outer check, but kept the one for the wizard as I didn't see a sanity check for that one. I had measured the position of the rear adjustment points on mine and got ~200, but I then did the math based on a 220mm bed size and assumed the others were offset from the edge by 20. I updated the pull request with your suggested values instead.

thisiskeithb:
> I removed the outer check, but kept the one for the wizard as I didn't see a sanity check for that one.

Sanity checks should be added upstream in Marlin instead of adding a one-off guard to this config.

thisiskeithb:
I just noticed that this PR does not target the `import-2.1.x` branch, so you'll need to retarget your changes against that branch.

cbagwell:
I like the idea of getting more balanced tramming points.  One concern is if you purchase the Creality CR-Touch upgrade kit, the mounting bracket will put you nozzle offsets around [-45, -7] so the X has biggest limit.  On my machine, that works out to 190 as max X probe point.  So I set my probe points at 45mm from each edge.  It's not on top of the screws but balanced and as close as I can get.

The current values are unbalanced but the 180 value is preventing a rash of build failure reports from ALIDATE_TRAMMING_POINT() for BL-Touch users since this area is a bit confusing to calculate good values for.

PR #873, opened 2022-12-31
Set TRAMMING_SCREW_THREAD to 40 for Ender 3 Printers

zanhecht:
### Requirements

### Description

Specify that the Ender 3 uses M4 bed screws.

### Benefits

Provides correct values for screw size if the ASSISTED_TRAMMING option is enabled.

### Related Issues

Updated version of #872, rebaselined to 2.1.x.

### DIFF
--- a/config/examples/Creality/Ender-3 Pro/BigTreeTech SKR Mini E3 2.0/Configuration_adv.h
+++ b/config/examples/Creality/Ender-3 Pro/BigTreeTech SKR Mini E3 2.0/Configuration_adv.h
@@ -1034,7 +1034,7 @@
 #if ENABLED(ASSISTED_TRAMMING)
 
   // Define positions for probe points.
-  #define TRAMMING_POINT_XY { {  20, 20 }, { 180,  20 }, { 180, 180 }, { 20, 180 } }
+  #define TRAMMING_POINT_XY { {  50, 50 }, { 185,  50 }, { 185, 185 }, { 50, 185 } }
 
   // Define position names for probe points.
   #define TRAMMING_POINT_NAME_1 "Front-Left"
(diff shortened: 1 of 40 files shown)


config/examples/Creality/Ender-3 Pro/BigTreeTech SKR Mini E3 2.0/Configuration_adv.h
```diff
@@ -1045,7 +1045,9 @@
   #define RESTORE_LEVELING_AFTER_G35    // Enable to restore leveling setup after operation
   //#define REPORT_TRAMMING_MM          // Report Z deviation (mm) for each point relative to the first
 
-  //#define ASSISTED_TRAMMING_WIZARD    // Add a Tramming Wizard to the LCD menu
+  #if EITHER(HAS_MARLINUI_MENU, HAS_TFT_LVGL_UI)
+    //#define ASSISTED_TRAMMING_WIZARD    // Add a Tramming Wizard to the LCD menu
+  #endif
```
thisiskeithb: To repeat what I said in https://github.com/MarlinFirmware/Configurations/pull/872, this change should not go here. Please add a sanity check upstream in Marlin to check for this invalid config instead of adding it to this single config.

config/examples/Creality/Ender-3 Pro/BigTreeTech SKR Mini E3 2.0/Configuration_adv.h
```diff
@@ -1034,7 +1034,7 @@
 #if ENABLED(ASSISTED_TRAMMING)
 
   // Define positions for probe points.
-  #define TRAMMING_POINT_XY { {  20, 20 }, { 180,  20 }, { 180, 180 }, { 20, 180 } }
+  #define TRAMMING_POINT_XY { {  50, 50 }, { 185,  50 }, { 185, 185 }, { 50, 185 } }
```
thisiskeithb: Since tramming points will depend on your probe type & placement relative to the nozzle, it's probably best to leave these as Marlin's defaults.

thinkyhead: We do want to tram where the screws are, if possible, so the default should probably go for those points, and then it can be adjusted for a probe that can't reach.

thisiskeithb: See my comment in https://github.com/MarlinFirmware/Configurations/pull/872#issuecomment-1368146126

Stock screw locations on Ender-3/5 beds are `TRAMMING_POINT_XY { {  32.5, 32.5 }, { 202.5,  32.5 }, { 202.5, 202.5 }, { 32.5, 202.5 } }`, but they may not be reachable depending on where the probe is located relative to the nozzle.

Ex: If your probe is to the left of the hotend, you may not be able to reach the screw heads on the right side of the bed or if your probe is to the right of the hotend, you may not be able to reach the screw heads on the left side of the bed.

It really depends on your probe/mount/probe offsets/x & y min & max positions.

zanhecht:
I'll leave the tramming points discussion for another time, and redefine this pull request to just changing the screw thread. I expanded it to all Ender 3 variants while I was at it.

PR #888, opened 2023-01-30
Single configuration #define for un-modified AnyCubic Mega Pro

joelpmichael:
### Requirements

Single configuration define for un-modified AnyCubic Mega Pro

### Description

The AnyCubic Mega Pro has 2 major differences from other printers in the i3 MEGA range - TMC2208 drivers on the X and Y axis, and a DDB geared extruder. A single define near the top of the configuration sets correct driver type and inversion, sets correct(ish) default E0 steps, and enables the DGUS-clone filename fix.


thinkyhead:
When it's not the "stock" configuration, and none of the other custom options are enabled, what is it? It seems to me that it should just be "stock" unless one of the custom options is enabled, so if it could be set up that way it would be clearer.

PR #889, opened 2023-01-31
Simax Mi-M200 initial configurations

kukumagi:
### Requirements

Simax Mi-M200 3d printer

### Description

Simax Mi-M200 3d printer configuration files.
Made by a beginner. With reference from ender 3 pro config.
Configurations might still be inaccurate but managed to print with my printer.

### Benefits

Adds configuration files for Simax Mi-M200

### Related Issues

--


ellensp:
You might want to update 
#define CONFIG_EXAMPLES_DIR "Creality/Ender-3 Pro/CrealityV422"

thisiskeithb:
> You might want to update
#define CONFIG_EXAMPLES_DIR "Creality/Ender-3 Pro/CrealityV422"

It should be removed since we have a script that adds that automatically to configs in non-import branches.

kukumagi:
> > You might want to update
> > #define CONFIG_EXAMPLES_DIR "Creality/Ender-3 Pro/CrealityV422"
> 
> It should be removed since we have a script that adds that automatically to configs in non-import branches.

Removed #define CONFIG_EXAMPLES_DIR "Creality/Ender-3 Pro/CrealityV422" from both config files

PR #890, opened 2023-02-01
Fix Artillery Genius Y_BED_SIZE

pierm8:
### Requirements

Artillery Genius

### Description

Changed Y_BED_SIZE value to 220 (previously 230) because the Artillery Genius has a bed size of 220x220.

### Benefits

Correct bed size.

### DIFF
--- a/config/examples/Artillery/Genius/BLTouch/Configuration.h
+++ b/config/examples/Artillery/Genius/BLTouch/Configuration.h
@@ -1757,7 +1757,7 @@
 
 // The size of the printable area
 #define X_BED_SIZE 220
-#define Y_BED_SIZE 230
+#define Y_BED_SIZE 220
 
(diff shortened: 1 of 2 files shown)


thisiskeithb:
The bed advertised as 220, but is really 230.

pierm8:
It is actually 230 however setting Y_BED_SIZE as 230 when printing the mesh validation pattern (G26) the print ends up right on the edge of the bed and makes a mess. I think it is more appropriate to set the size to 220x220 as it is also in the original firmware.

thisiskeithb:
> It is actually 230 however setting Y_BED_SIZE as 230 when printing the mesh validation pattern (G26) the print ends up right on the edge of the bed and makes a mess.

That should be fixed with `X/Y_MIN/MAX_POS` instead.

pierm8:
Ok, I'll try to adjust those parameters.

PR #897, opened 2023-02-17
Corrects the default leveling bed temperature (Ender-3 Neo)

jamespearson04:
### Description

 Corrects the default leveling bed temperature "LEVELING_BED_TEMP" from 150C to 50C in the example configuration for the Ender-3 Neo.

### Benefits

Untested, but presumably would have failed to compile or prevented bed leveling from working, as BED_MAXTEMP is 120C. Should work as expected.


thisiskeithb:
Thanks for your contribution! Unfortunately we can't accept PRs directed at bugfix branches. We make patches to the import branches and only later do we push them out to bugfix / as releases. Please redo this PR starting with the `import-2.1.x` branch and be careful to target `import-2.1.x` when resubmitting the PR.

ellensp:
Since it takes but a minute, I have created an identical PR in the import branch.  
This will be used to regenerate the other versions of the config

jamespearson04:
I see, my bad. Many thanks!

PR #902, opened 2023-03-07
Support for Tronxy XY-2 Pro

maksymovi:
### Description

Added configuration for the Tronxy XY-2 Pro. 

Configuration currently is for a machine modified with a BLTouch. I make this clear in the Configuation.h file, and have instructions on how to enable the stock inductive sensor, kept it this way as I have not tested it this way.

### DIFF
--- a/config/examples/Tronxy/XY2-Pro/Configuration.h
+++ b/config/examples/Tronxy/XY2-Pro/Configuration.h
@@ -1,0 +1,3499 @@
+/**
+ * Marlin 3D Printer Firmware
+ * Copyright (c) 2020 MarlinFirmware [https://github.com/MarlinFirmware/Marlin]
+ *
+ * Based on Sprinter and grbl.
+ * Copyright (c) 2011 Camiel Gubbels / Erik van der Zalm
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+#pragma once
+
+/**
+ * Configuration.h
+ *
+ * Basic settings such as:
+ *
+ * - Type of electronics
+ * - Type of temperature sensor
+ * - Printer geometry
+ * - Endstop configuration
+ * - LCD controller
+ * - Extra features
+ *
+ * Advanced settings can be found in Configuration_adv.h
+ */
+#define CONFIGURATION_H_VERSION 02010300
+
+//===========================================================================
+//============================= Getting Started =============================
+//===========================================================================
+
+/**
+ * Here are some useful links to help get your machine configured and calibrated:
+ *
+ * Example Configs:     https://github.com/MarlinFirmware/Configurations/branches/all
+ *
+ * Průša Calculator:    https://blog.prusaprinters.org/calculator_3416/
+ *
+ * Calibration Guides:  https://reprap.org/wiki/Calibration
+ *                      https://reprap.org/wiki/Triffid_Hunter%27s_Calibration_Guide
+ *                      https://sites.google.com/site/repraplogphase/calibration-of-your-reprap
+ *                      https://youtu.be/wAL9d7FgInk
+ *
+ * Calibration Objects: https://www.thingiverse.com/thing:5573
+ *                      https://www.thingiverse.com/thing:1278865
+ */
+
+// @section info
+
+/**
+ * *** IMPORTANT INFO FOR THE TRONXY XY2-PRO PLEASE READ ***
+ * 
+ * This config is for the TronXY XY2-Pro Single Titan Extruder.
+ * So far it has been tested on a single machine (mine).
+ * It works on mine, may or may not work on yours, use at your own risk.
+ * It similarly may or may not work on other XY2-Pro models.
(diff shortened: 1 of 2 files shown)


thisiskeithb:
> I make this clear in the Configuation.h file, and have instructions on how to enable the stock inductive sensor

The configuration should be changed to use the stock probe with an option/flag at the top for a BLTouch, not the other way around.

maksymovi:
> > I make this clear in the Configuation.h file, and have instructions on how to enable the stock inductive sensor
> 
> The configuration should be changed to use the stock probe with an option/flag at the top for a BLTouch, not the other way around.

I can change it to what I believe the right config to be for this but cannot test it given I don't have the probe. Would that be fine?

thisiskeithb:
That works

maksymovi:
Updated to fix this.

g4m3r0:
The Z-homing does not work for me with the latest bugfix-2.1.x branch. It goes to the middle of the bed, then goes up two times and the z-homing is finished. It does not go down at all.

I've got a unmodified Troxy XY-2 Pro with the CXY-V6-191017 board. 
Do you have any idea what could be wrong with the config or my environment?

thisiskeithb:
> The Z-homing does not work for me with the latest bugfix-2.1.x branch. It goes to the middle of the bed, then goes up two times and the z-homing is finished. It does not go down at all.
> 
> I've got a unmodified Troxy XY-2 Pro with the CXY-V6-191017 board. Do you have any idea what could be wrong with the config or my environment?

You likely need to change `Z_MIN_ENDSTOP_HIT_STATE` & `Z_MIN_PROBE_ENDSTOP_HIT_STATE` from `HIGH` to `LOW`.

g4m3r0:
> You likely need to change `Z_MIN_ENDSTOP_HIT_STATE` & `Z_MIN_PROBE_ENDSTOP_HIT_STATE` from `HIGH` to `LOW`.

Thanks for your reply! This actually was the first thing I've tried. But it resulted in the nozzle crashing into the heatbed.
I do use the CHITU3D B6 board and the chitu_f103 environment.

thinkyhead:
It's possible the wrong probe pin is assigned, or the probe is connected to a different pin. Can you double check?

g4m3r0:
@thinkyhead its all hard wired on that CXY-V6-191017 board. 
Previously I had flashed an older marlin version from this fork https://github.com/rhapsodyv/Marlin , where the probe was working fine. But I had problems with crashes every now and then.

According to this schematics the probe pin matches the CHITU3D B6 configuration (10kΩ Pullup to +5V and 100nF to GND, then 20kΩ to MCU Pin124 PG9).
https://static.wikia.nocookie.net/tronxy/images/7/76/CXY-V6-19107_pinout_and_description.pdf/revision/latest?cb=20240502234640

I found this two issues with a similar issue:
https://github.com/MarlinFirmware/Marlin/issues/21339
https://github.com/MarlinFirmware/Marlin/issues/23677

I have tried to add the following suggested answers to the marlins setup method, but without any luck.
`
  SET_INPUT_PULLDOWN(Z_MIN_PIN);
  SET_INPUT_PULLUP(Z_MIN_PIN);
`
or
`GPIOA->BSRR = (1U << PG9);`

I've also tried to uncomment
`
#define ENDSTOPPULLUP_ZMIN
#define ENDSTOPPULLUP_ZMIN_PROBE
`

thomas-haller:
I had to do the following modifications to make it running with my Tronxy XY2 Pro (no modifications):
```
// for the default z probe to work:
#define Z_MIN_PROBE_ENDSTOP_HIT_STATE LOW
#undef Z_MIN_PROBE_USES_Z_MIN_ENDSTOP_PIN
#define USE_PROBE_FOR_Z_HOMING

// for the default extruder to work:
#define DEFAULT_AXIS_STEPS_PER_UNIT   { 160, 160, 800, 186 } //values taken from https://github.com/rhapsodyv/Marlin/
#define INVERT_E0_DIR false

```

thinkyhead:
So are the corrections proposed by @thomas-haller accurate? I will be happy to apply the changes if no one else is in the mood to do so.

thinkyhead:
See 9e7af8169d

thomas-haller:
I didn't expect it to get merged :grinning: 
Else I would have opened a new issue

I just found yesterday that you need
`
#define E0_AUTO_FAN_PIN FAN2_PIN
`
in Configuration_adv.h to get the extruder fan working

BeowulfNode42:
> I just found yesterday that you need `#define E0_AUTO_FAN_PIN FAN2_PIN` in Configuration_adv.h to get the extruder fan working

The config that works on my CXY-V6-191017 board in my X5SA Pro also needs that line in its configuration_adv.h. Then in the pins_CHITU3D_common.h the existing value for `#define FAN2_PIN PG14` works fine.

I keep trying to refer to the code in https://github.com/tronxy3d/F4xx-SIM480x320 for this board, but it's not easy as its old Marlin and they've replaced large sections with files in strange places and things are sometimes called different things. However to get the Z axis homing I had to uncomment `//#define PINS_DEBUGGING` in configuration_adv.h so I could use [pronterface](https://github.com/kliment/Printrun) and the gcode [`M43`](https://marlinfw.org/docs/gcode/M043.html) as the `M119` gcode command initially showed no change and no hit state for my Z-axis inductive probe triggered or not.

If you're still having troubles with your Z-axis homing, check with the `M119` and `M43` gcode commands to see what's happening.

thinkyhead:
> The config that works on my CXY-V6-191017 board in my X5SA Pro also needs that line in its configuration_adv.h. 

And it is already so in the X5SA Pro example configuration… which should be merged pretty soon. If it looks good to you.

> but it's not easy as its old Marlin and they've replaced large sections

Comparing it to the main Marlin fork version that it's based on makes the changes pretty obvious. The date in the `Version.h` file is one of the better clues where to start from. So, you would first do a checkout of the main fork at that point in time, and then you would delete all the files (except the `.git` folder) and drop in the modified files from the tronxy3d fork. Viewing this in a tool like GitHub Desktop makes clear all the changes, and then those can be manually applied and adapted to the latest codebase.

PR #904, opened 2023-03-10
DGUS_RELOADED 

Neo2003:
#### DWIN_SET version 1.0.3

- Z is now displayed with 2 decimals and rounded to 0.02 (was one decimal)
-  Bed temperature from int to float with 1 decimal
-  Nozzle temperature from int to float with 1 decimal
-  Screens adjusted accordingly with type changes and more room

### Requirements

Requires the pull request #25490 in 2.1_bugfix branch of MarlinFirmwares https://github.com/MarlinFirmware/Marlin/pull/25490
Please see more details and pictures in the above PR.

### Description

This refines the Z displayed in Printing screen by showing 2 decimals instead of only one. The value is rounder to 0.02.
This refines current bed and nozzle temperature displayed in all pages by changing the type from integer to float with one decimal.

### Benefits

More fine values displayed in order to view ABL working by seeing the changing Z, also temperatures are a lot better with one decimal to check for example how PID is acting.

### Related Issues

N/A


thinkyhead:
I added the following files to the `DWIN_SET` because they are not otherwise included:

- `T5UID1_V30.BIN`
- `T5OS_V21_NOACK.BIN`

Are these files both required, or is only one of them required for the specific screen, or is neither required because the screen will retain the OS it already has installed?

InsanityAutomation:
The OS flash is dangerous, and if coming from a different major rev will cause screens to lose touch response until reflashed with a config to perform calibration. Also that OS rev is quite old and wont function on newer screens.

Noack is fine.

thinkyhead:
> Noack is fine.

@InsanityAutomation — So, it would be best to leave the NOACK file in there and remove the other…? Should we preserve the `T5UID1_V30.BIN` file someplace, in case it gets removed from its current home repo, or is it truly obsolete?

InsanityAutomation:
Truly obsolete in about the same sense as trying to install Windows XP on something. You can safely update any of the 3.0 or 3.5 screens over to 4.0 or 4.5. I currently recommend the 4.0 with 50 millisecond response. 

I'm not even sure that the noack file is even still processed by current firmware revisions to be honest but it doesn't cause any harm.

thinkyhead:
> I currently recommend the 4.0 with 50 millisecond response.

@InsanityAutomation — How would you recommend we provide this version to users? I want to make sure they don't have to go too far afield to get all the resources they need to keep their LCD up to date.

InsanityAutomation:
It's more of a get it if you need it, build with the proper dgus tool and it works with OS2 through 4.5 seamlessly. Only times I hear issues is when they try to update the OS without the calibration config, rarely hear of issues if they don't touch the os.

thinkyhead:
@InsanityAutomation — So in short, I should remove the BIN files and just keep the assets, and users will not have any issues, correct?

I may be a little selfish in that I also want to keep the best BIN file around somewhere I can easily find it and share it with Marlin collaborators. With all the things I have to track I'm always a bit afraid that months down the line I'll be unable to find an obscure link that I saved in a note or browser.

thinkyhead:
This isn't quite ready to be merged. The configuration for the "stock" Ender 5 Plus with creality V1 board uses the `DGUS_LCD_UI ORIGIN` setting. However the README is only referencing the `RELOADED` version of the UI. The README should be updated with the correct instructions to use `DGUS_LCD_UI ORIGIN` if that is the preferred "stock" UI. Or, the configuration should be changed to use `DGUS_LCD_UI RELOADED` if the `ORIGIN` user interface is substandard. Ideally, the README should contain instructions for all the UI options that can be used with the stock display.

PR #922, opened 2023-04-15
ABS preheat value fixed

brandon-northcutt:
### Requirements

### Description

This PR changes the ABS preheat temperature to avoid the assert causing the following error.

Marlin\src\module\temperature.cpp:335:38: error: static assertion failed: PREHEAT_2_TEMP_HOTEND (245) must be less than HEATER_0_MAXTEMP (250) - 15.

As ABS melts between 210 and 250, 210 is a better preheat temperature anyway.

### Benefits

Current config will not compile in bugfix-2.1.x because of the above assert in temperature.cpp.

### Related Issues
N/A

### DIFF
--- a/config/examples/Creality/CR-10S/CrealityV1/Configuration.h
+++ b/config/examples/Creality/CR-10S/CrealityV1/Configuration.h
@@ -2356,7 +2356,7 @@
 #define PREHEAT_1_FAN_SPEED     0 // Value from 0 to 255
 
 #define PREHEAT_2_LABEL       "ABS"
-#define PREHEAT_2_TEMP_HOTEND 250
+#define PREHEAT_2_TEMP_HOTEND 210 // ABS working temp between 210 and 250 C.
 #define PREHEAT_2_TEMP_BED    80
 #define PREHEAT_2_TEMP_CHAMBER 35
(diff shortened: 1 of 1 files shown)


config/examples/Creality/CR-10S/CrealityV1/Configuration.h
```diff
@@ -2356,7 +2356,7 @@
 #define PREHEAT_1_FAN_SPEED     0 // Value from 0 to 255
 
 #define PREHEAT_2_LABEL       "ABS"
-#define PREHEAT_2_TEMP_HOTEND 250
+#define PREHEAT_2_TEMP_HOTEND 210 // ABS working temp between 210 and 250 C.
```
thisiskeithb: I’d recommend dropping this to 235 to pass the max temperature check instead of dropping it to 210.

brandon-northcutt: Sounds good to me.

thinkyhead: 210 + 15 = 225 … so it can be reduced by 10°C if preferred.

Issue #823, opened 2022-10-06
ENDSTOP_INTERRRUTS break blouch on creality boards, possibly other STM ones


EvilGremlin:
z_min_probe is stuck triggered with interripts, i think it's better to disable them at least for creality boards

thisiskeithb:
How many reports have we seen with this issue? Were they using genuine BLTouches or clones?

EvilGremlin:
Couple so far, one in dicsord, other in local forum. Clone or not doesnt matter - it fails in force_sw too. Fixing configs should be stopgap until someone get to actual issue

ellensp:
From a user that just disabled ENDSTOP_INTERRRUTS and it started to work
"its a ender 3 pro, stock mobo, 4.2.2, 256k i think? though it may be 512k but im flashing with 256k anyways because i got it recently. BLTouch is plugged into the BL Touch port on the mobo"
BLTOUCH looks genuine

thisiskeithb:
> it fails in force_sw too

Something is really broken if `BLTOUCH_FORCE_SW_MODE` is now causing BLTouches to fail.

EvilGremlin:
??? 
it doesn't

silycr:
Using Z_Min Endstop with 3dTouch on Tronxy X5S-2e stock board (Chitu cxy-v5-180409) STM32F1. Works both EN and DISABLED ENDSTOP_INTERRRUPTS. 
EN BLTOUCH_FORCE_SW_MODE can not detect 3dTouch. DISABLED can be sensed. Possible 3DTouch issue. Will test with genuine BLTouch Classic.

bellertm1:
Hardware: Ender3 Pro and 5, Official Creality BLTouch and STM32F103RE 4.2.7 board
Software: tried 2.1.X Bugfix,   Release 2.1.1,  and 2.0.9.5 all with 512 Flash 
Settings:
#define Z_MIN_PROBE_USES_Z_MIN_ENDSTOP_PIN
AND
// Force the use of the probe for Z-axis homing
#define USE_PROBE_FOR_Z_HOMING
AND
 * The BLTouch probe uses a Hall effect sensor and emulates a servo.
 */
#define BLTOUCH

[...]

PR #924, opened 2023-04-22
Improve Ender-3 with SKR Mini E3 V3 Config

thisiskeithb:
### Description

Improve Ender-3 with SKR Mini E3 V3 Config

### Benefits

Bring config more in line with other SKR Mini E3 configs

### DIFF
--- a/config/examples/Creality/Ender-3/BigTreeTech SKR Mini E3 3.0/Configuration.h
+++ b/config/examples/Creality/Ender-3/BigTreeTech SKR Mini E3 3.0/Configuration.h
@@ -62,7 +62,7 @@
 // @section info
 
 // Author info of this build printed to the host during boot and M115
-#define STRING_CONFIG_H_AUTHOR "(BigTreeTech, SKR-mini-E3-V3.0)" // Who made the changes.
+#define STRING_CONFIG_H_AUTHOR "(BigTreeTech, Ender-3)" // Who made the changes.
 //#define CUSTOM_VERSION_FILE Version.h // Path from the root directory (no quotes)
 
 // @section machine
@@ -408,7 +408,7 @@
   //#define MKS_PWC                 // Using the MKS PWC add-on
   //#define PS_OFF_CONFIRM          // Confirm dialog when power off
   //#define PS_OFF_SOUND            // Beep 1s when power off
-  #define PSU_ACTIVE_STATE HIGH      // Set 'LOW' for ATX, 'HIGH' for X-Box
+  #define PSU_ACTIVE_STATE LOW      // Set 'LOW' for ATX, 'HIGH' for X-Box
 
   //#define PSU_DEFAULT_OFF               // Keep power off until enabled directly with M80
   //#define PSU_POWERUP_DELAY      250    // (ms) Delay for the PSU to warm up to full power
@@ -642,7 +642,7 @@
 #define HEATER_5_MAXTEMP 275
 #define HEATER_6_MAXTEMP 275
 #define HEATER_7_MAXTEMP 275
-#define BED_MAXTEMP      150
+#define BED_MAXTEMP      125
 #define CHAMBER_MAXTEMP  60
 
 /**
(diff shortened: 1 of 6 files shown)


config/examples/Creality/Ender-3/BigTreeTech SKR Mini E3 3.0/Configuration.h
```diff
@@ -1173,7 +1172,7 @@
 
 // Enable this feature if all enabled endstop pins are interrupt-capable.
 // This will remove the need to poll the interrupt pins, saving many CPU cycles.
-#define ENDSTOP_INTERRUPTS_FEATURE
+//#define ENDSTOP_INTERRUPTS_FEATURE
```
thinkyhead: If this can be enabled, it's usually helpful. AFAIK all ARM board endstop pins should be interrupt-capable.

config/examples/Creality/Ender-3/BigTreeTech SKR Mini E3 3.0/Configuration_adv.h
```diff
@@ -1779,7 +1779,7 @@
 
   //#define SD_REPRINT_LAST_SELECTED_FILE // On print completion open the LCD Menu and select the same file
 
-  #define AUTO_REPORT_SD_STATUS           // Auto-report media status with 'M27 S<seconds>'
+  //#define AUTO_REPORT_SD_STATUS         // Auto-report media status with 'M27 S<seconds>'
```
thinkyhead: This is another one of those options that can help reduce overhead, so no harm in keeping it enabled if it fits.

config/examples/Creality/Ender-3/BigTreeTech SKR Mini E3 3.0/Configuration.h
thisiskeithb: > If this can be enabled, it's usually helpful. AFAIK all ARM board endstop pins should be interrupt-capable.

See https://github.com/MarlinFirmware/Configurations/issues/823

thinkyhead: This could be due to an uncaught conflict between interrupt-enabled pins and related timers or ISRs. It would be worth an investigation by checking the STM32 datasheets to see if the problematic endstop pins coincide with timers or ISRs and whether we can change how we initialize the related pins, timers, and ISRs.

cbagwell:
I aligned my config file with these changes and tested them and all went fine.  A few notes.

* These changes apply to Ender 3 Pro config in case you want to update that set as well (what I tested).
* I left endstop interrupt enabled because its worked well for me so far.
* I didn't test BLTOUCH changes since mine is connected to 5-pin connector.  Perhaps that reduced my risk of misconfiguring BLTOUCH and leading to endstop issue as mentioned in that other issue.
* I didn't test the PID values.

thinkyhead:
> These changes apply to Ender 3 Pro config

@cbagwell — Every change without exception? What is the primary difference with the Ender-3 Pro over the Ender-3?

EvilGremlin:
Better PSU

thisiskeithb:
> What is the primary difference with the Ender-3 Pro over the Ender-3?

As far as Marlin is concerned? Basically nothing.

Pro configs say “Ender-3 Pro” on the status screen “Ready” message and `_Statusscreen.h`.

cbagwell:
I'd almost forgotten about that import status screen. 😄  The Pro also comes with a magnetic bed instead of using clips but there are no features enabled that require knowledge of that (no probe for example).  And I noticed even non-Pro is starting to ship with magnetic beds depending where you purchase and I wouldn't be surprised if they are totally the same as Pro's now since even the Pro no longer ships with a Meanwell PSU.

thisiskeithb:
> I wouldn't be surprised if they are totally the same as Pro's now since even the Pro no longer ships with a Meanwell PSU.

The lines between Ender-3 SKUs are quite blurred at this point.

thisiskeithb:
We'd have a lot more Ender-3 Pro configs if we:

- Back up the Ender-3 Pro's `_Statusscreen.h`
- Nuke the contents of the Ender-3 Pro configs folder
- Copy over Ender-3 configs & replace previously backed up `_Statusscreen.h`
- Replace all mentions of "Ender-3" with "Ender-3 Pro"

...Then remember to always add an Ender-3 Pro config for each new Ender-3 config. Or automate it.

thinkyhead:
> We'd have a lot more Ender-3 Pro configs…

Sounds like all that could be done with a README that explains the Ender-3 Pro is just the Ender-3 with some settings and files adjusted. But, if all Ender-3 configs were ported over to Ender-3 Pro configs it wouldn't be a major addition to the existing maintenance burden, as we're already doing "mostly automated" maintenance of over 300 configs now.

PR #934, opened 2023-05-14
TronXY X5SA-500 Pro config

lukasradek:
### Requirements

I have added example config for TronXY X5SA-500 Pro.


thinkyhead:
ICYMI — This was closed by the robot because PRs must be targeted at `import-2.1.x` because other branches are generated by automated scripts.

PR #936, opened 2023-05-26
Tarantula Pro with MKS SGEN-L

thisiskeithb:
### Description

The Tarantula Pro originally shipped with an MKS GEN-L (8-bit) and was later shipped with an SGEN-L (32-bit), so I've added that option.

This PR also:
- Generalizes extruder auto fan to use `FAN1_PIN` which will work correctly for both the GEN-L and SGEN-L builds
- Enables Mesh Bed Leveling since there's plenty of room to spare & these beds are not flat 😄 

### Benefits

Improved config for both new & old Tarantula Pro builds.

### Related Issues

None. Came up on Discord.


config/examples/Tevo/Tarantula Pro/Configuration.h
```diff
@@ -23,6 +23,8 @@
 #error "Don't build with import-2.1.x configurations!"
 #error "Use the 'bugfix...' or 'release...' configurations matching your Marlin version."
 
+//#define MOTHERBOARD BOARD_MKS_SGEN_L  // Uncomment for the MKS SGEN-L board (32-bit), otherwise MKS GEN-L board (8-bit) is assumed
```
thinkyhead, line 26: I wish we had a way to do this that was friendlier for the Custom Configuration Editor in Auto Build Marlin. We'll have to hire a board of consultants and a staff of Ph.D. programmers to sort this one out.

thisiskeithb: Indeed. True config parsing would be slick, especially for forks I maintain that support multiple boards.

thinkyhead:
I went ahead and split the config in two since the 32-bit board is likely to be able to support more features, so over time it can get more enhancements added.

thisiskeithb:
> I went ahead and split the config in two since the 32-bit board is likely to be able to support more features, so over time it can get more enhancements added.

Just so I'm clear, is this only for machines with both 8-bit & 32-bit options?

I've combined multi-board configs into single configs for the Biqu B1 & (separately) BX, so those should stay combined as they are 32-bit builds?

thinkyhead:
> Just so I'm clear, is this only for machines with both 8-bit & 32-bit options?

Any case where the `MOTHERBOARD` differs enough that it can support (some… many) more added features.

There's a general question as to how important the new tools I'm building for user configuration may become in the future, and whether we end up putting all the configurations into a database so we can auto-generate them, and whether the tools I'm building can be adaptable to configurations that have added options and conditions, and so on….

There is a growing tension between keeping the number of configuration files to a minimum versus these other considerations, especially as the number of configurations approaches 400 and beyond.

PR #737, opened 2022-05-16
Fixes from "Build All Examples" Script

thisiskeithb:
### Description

Fix up configs so they are buildable using Marlin's `buildroot/bin/build_all_examples` test script.

This is one giant commit, but I can break it up into several commits/descriptions if you have questions (or just comment on the line(s).)

### Benefits

Most configs are buildable now! 

The outstanding / unbuildable configs are listed below. Also, some of those require fixes in the main Marin repo.

PR #937, opened 2023-05-26
Guard Some LCD_SERIAL_PORT Defines

thisiskeithb:
### Description

Prevent defining `LCD_SERIAL_PORT` if user ports config to different hardware / chooses not to use stock LCD.

I didn't update the Ender-5 S1 config since that seems like a can of worms with all kinds of LCD-specific settings.


thisiskeithb:
> `#error "Serial Port 1 pin D18 and/or D19 conflicts with another pin on the board."`

Well, that isn't my fault 😝

thinkyhead:
> Well, that isn't my fault 😝

Might be the first time this config has been built through the Configurations PR testing. I just tried building the Malyan M200 config with current bugfix-2.1.x and it was fine.

thisiskeithb:
> Might be the first time this config has been built through the Configurations PR testing. I just tried building the Malyan M200 config with current bugfix-2.1.x and it was fine.

It's actually [the Nextion config that is the issue](https://github.com/MarlinFirmware/Configurations/actions/runs/5095276906/jobs/9160060143#step:9:110). I found it last year as well: https://github.com/MarlinFirmware/Configurations/pull/737#issuecomment-1128308417

It's an impossible config since since Z endstops conflict with serial port 1.

thinkyhead:
Changed Nextion example to serial port 2.

thisiskeithb:
Now the Malyan M300 config is failing due to unreachable probe points, but as I also noted in https://github.com/MarlinFirmware/Configurations/pull/737#issuecomment-1128308417, I don't know what the proper margins/offsets/values are for this machine since I don't own one to test on.

PR #938, opened 2023-05-28
Update Configuration_adv.h

TherealJoeyw:
### Requirements

### Description

added a tag that because absent would cause `#error "Your Configuration_adv.h file is for an old version of Marlin. Downgrade Marlin or upgrade your Configuration_adv.h` to appear and cause build to fail

### Benefits

allows the file to serve its primary funtion

### DIFF
--- a/config/examples/JGAurora/A5/Configuration_adv.h
+++ b/config/examples/JGAurora/A5/Configuration_adv.h
@@ -21,6 +21,8 @@
  */
 #pragma once
 
+#define CONFIG_EXAMPLES_DIR "JGAurora/A5"
+
 /**
  * Configuration_adv.h
  *


thisiskeithb:
> #error "Your Configuration_adv.h file is for an old version of Marlin. Downgrade Marlin or upgrade your Configuration_adv.h 

This error occurs because of mismatched versions of Marlin (i.e: `CONFIGURATION_H_VERSION` / `CONFIGURATION_ADV_H_VERSION` used with the incorrect version of Marlin from upstream), not a missing `CONFIG_EXAMPLES_DIR` flag. In fact, `CONFIG_EXAMPLES_DIR` is not checked anywhere within Marlin's code and is only used to identify where the config came from.

Also, the `CONFIG_EXAMPLES_DIR` flag is not used in the `import-2.*` branches and is automatically added to the `bugfix-2.*` and `release-2.*` branches which is already there for this config (and all others):

https://github.com/MarlinFirmware/Configurations/blob/95dfab0b426b3a51946c8b25358615b0c15b5258/config/examples/JGAurora/A5/Configuration.h#L24

PR #942, opened 2023-05-30
Add support for CRTouch to Ender-3 Pro 4.2.7 board

trickyearlobe:
### Description

Hi guys, I created this template as there seems to be a lot of FUD circulating on the internet about how exactly to get a reliable Ender3 configuration with a working CR-Touch and how to level it properly.

This seems to stem from :-
* The stock Creality firmwares are woefully out of date/neglected
* The stock Creality firmwares seem to be misconfigured in weird ways
* There are several conflicting guides about how to compile your own which reference old (refactored) config params

The version in this PR is compiled from various technical scraps I pulled together from several sources and seems pretty reliable for me when used to compile the latest (2.1.2.1) release so I thought I'd offer it to the repo

**If it's policy not to accept "not-stock" configs, I totally understand, but I thought it would be nice for people to get a head start and not have to scour forums to get something working easily. While this config may not be perfect, its seems (to me) to be a reliable starting point for Ender-3 Pro with CR-Touch installed. I'd love for people to iterate on it to get something better than the Creality stock version.**

### Benefits

Easy auto homing with CR-Touch using the homing/levelling wizard (thank you for that marvel... its just so easy to use)

Fixes all my issues with:-
* Levelling the bed
  * forgotten settings
  * inability to set the Z offset below zero on some firmwares
  * crashing head into the bed on some firmwares
* Prints starting in midair when printing consecutively
* Extruder stopping half way through a print job, but printer continues executing the moves.


thisiskeithb:
Thanks for your contribution! Unfortunately we can't accept PRs directed at release branches. We make patches to the import branches and only later do we push them out as releases. Please redo this PR starting with the `import-2.1.x` branch and be careful to target `import-2.1.x` when resubmitting the PR.

EvilGremlin:
We don't leave bltouch (or runout) enabled in example configs if it doesnt come with stock machine. Common way is to add something like `#define HAS_BLTOUCH` at the top and use it to enable all relevant config parts.

trickyearlobe:
Hi @thisiskeithb , thanks for the info on your PR process.... totally makes sense.

Hi @EvilGremlin , if I understood correctly, you'd recommend adding a define and some conditionals to the existing example config for Ender3 Pro 4.2.7 board ?

I'm totally fine with reworking the PR for that, but I wonder if it would be more "DRY" to do that elsewhere (such as in the in the firmware repo) so that all the clones (which also have example configs) can benefit. It seems sensible for instance to turn on Z homing safety and the Z wizard for all models if BLTouch is enabled.

thisiskeithb:
Note: A BLTouch option has been added in the past & removed since these machines don't ship with one:
- https://github.com/MarlinFirmware/Configurations/pull/895

> It seems sensible for instance to turn on Z homing safety and the Z wizard for all models if BLTouch is enabled.

`Z_SAFE_HOMING` is sensible, but setting other options automatically can increase the default build size & isn't always wanted.

cbagwell:
If HAS_* is deemed acceptable then HAS_CRTOUCH might be better name since its currently not a generic term like BLTOUCH and lets us assume important bits based the CR-TOUCH kit that Creality markets for their Enders.

* Kit ships with 5 pin connector so we know for sure to disable Z_MIN_PROBE_USES_Z_MIN_ENDSTOP_PIN and enable USE_PROBE_FOR_Z_HOMING.
* Kit ships with a mount of known size and CR-Touch is of known size so we can provide a meaningful NOZZLE_TO_PROBE_OFFSET.
* Creality markets the kit for Ender 3 (and Pro/V2/Max versions) and Ender 5 (and S/Pro versions) and CR-10 as well as the v4.2.2 and v4.2.7 boards those same ship with.
    * Thats like 16 config files that could use HAS_CRTOUCH... not to mention at least the drop-in replacements (SKR Mini E3 V1/2/3) could use it as well.

On the other hand, a HAS_BLTOUCH doesn't help much with probe pins (5 pin cable) vs. endstops (3+2 cable) config values or probe offset since it could mean any generic BLTOUCH and custom printed probe mount combo.  

Also, I would guess CRTOUCH kit is not very popular outside of targeted Ender's since its a bit overpriced for machines that can't leverage the decent metal mount.

trickyearlobe:
I'd add that #895 linked by @thisiskeithb is not just defaulting the device to disabled... it's removing the `#define` and the conditionals that reference it. That suggests to me that `#define HAS_*` may not actually be acceptable.

cbagwell:
I wonder if it was more a formatting issue that was the concern.  Instead of wrapping each standard definition with an #if; which could make future mass updates more complicated; how about a single block somewhere and leave existing values commented out?

```
//#define HAS_CRTOUCH
#if HAS_CRTOUCH
  #define BLTOUCH
  #define AUTO_BED_LEVELING_BILINEAR
  ...
#endif
```

Either way though I could see how it might become a maintenance headache if config files have unique #if structures.

trickyearlobe:
For now I'll just wait and see if the maintainers offer some consensus on how it should be structured. Not super keen to spend time on a lot of PR reworks without some confidence it would get merged.

PR #946, opened 2023-06-05
Creality Ender-3 S1 Plus / Pro printers example configurations

strnk:
### Requirements

MarlinFirmware/Marlin#25905 should be merged beforehand to support the touchscreen used on Ender-3 S1 Plus and Pro printers.

### Description

 Example configurations for the Ender-3 S1 Plus and Pro printers.

### DIFF
--- a/config/examples/Creality/Ender-3 S1 Plus/Configuration.h
+++ b/config/examples/Creality/Ender-3 S1 Plus/Configuration.h
@@ -634,7 +634,7 @@
 // Above this temperature the heater will be switched off.
 // This can protect components from overheating, but NOT from shorts and failures.
 // (Use MINTEMP for thermistor short/failure protection.)
-#define HEATER_0_MAXTEMP 320
+#define HEATER_0_MAXTEMP 275
 #define HEATER_1_MAXTEMP 275
 #define HEATER_2_MAXTEMP 275
 #define HEATER_3_MAXTEMP 275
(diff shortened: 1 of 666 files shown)


config/examples/Creality/Ender-3 S1 Plus/Configuration.h
```diff
@@ -0,0 +1,3559 @@
...
+  //#define FIL_RUNOUT7_PULLDOWN
+
+  //#define FIL_RUNOUT8_STATE LOW
+  //#define FIL_RUNOUT8_PULLUP
+  //#define FIL_RUNOUT8_PULLDOWN
+
+  // Commands to execute on filament runout.
+  // With multiple runout sensors use the %c placeholder for the current tool in commands (e.g., "M600 T%c")
+  // NOTE: After 'M412 H1' the host handles filament runout and this script does not apply.
+  #define FILAMENT_RUNOUT_SCRIPT ""
```
thinkyhead: What effect are we trying to get with an empty `FILAMENT_RUNOUT_SCRIPT`? This is not formally supported, so if some special behavior is being gained here, we should aim to get it by having `FILAMENT_RUNOUT_SCRIPT` commented-out instead.

strnk: No special behavior, I did not write the support of ADVANCED_PAUSE_FEATURE with the touch-screen yet, so I removed M600 in the default script without thinking of commenting it out.

strnk:
It turns out that undefined FILAMENT_RUNOUT_SCRIPT is not supported if filament sensor is enabled ATM. I fixed it in MarlinFirmware/Marlin#25905

PR #949, opened 2023-06-15
Ender 3V2 Neo Change XY Size and, MAX Temp E0 and BED, Grid BL Touch and Speed Calibration

DaniloRibeiro07:
The Ender3V2Neo has build volume: 220x220x250 mm, E0 temp max 260 °C and Hot Bed 100°C.

By: https://www.creality.com/br/products/ender-3-v2-neo-3d-printer

With this change, the bed and hot end don't colision with the struture.


thisiskeithb:
> The Ender3V2Neo has build volume: 220x220x250

What are the actual bed measurements / travel limits? Creality markets most Enders with a 220x220mm build volume, but they’re 235mm.

DaniloRibeiro07:
You are right.
However, when I put the dimensions of the table in configuration.h, it crashes with the structure. I'll do a more accurate analysis of the problem.

PR #968, opened 2023-08-26
use DWIN_MARLINUI_PORTRAIT on ender3 v2 marlin user interface. 

ellensp:
Ender 3 V2 screens are by default in portrait mode not landscape. So the Marlin UI default should be portrait

### Requirements

Ender 3 v2 display with encoder running marlin ui

### Description

Ender 3 V2 screens are by default in portrait mode not landscape. So the Marlin UI default should be in portrait

### Benefits

Less confused users

### Related Issues

https://github.com/MarlinFirmware/Marlin/issues/26205


ellensp:
Test fails on: "#error "CAUTION! Ender-3 V2 display requires a custom cable with TX = P0_15, RX = P0_16. See 'pins_BTT_SKR_E3_TURBO.h' for details. (Define NO_CONTROLLER_CUSTOM_WIRING_WARNING to suppress this warning.)""

Which is deliberate. We want this to error when a user tries this config

PR #981, opened 2023-10-20
Updated CR-10 Max Configuration.h

DakotaCodes00:
### Description

For some odd reason the CR-10 Max uses pin 19 for the BL Touch sensor pin and not the default Z- endstop pin or Z Probe Pin as defined in the ramps pin file.

Previously used this config and the BL Touch never triggered, explicitly using pin 19 fixed the issue.

I also just changed the machine name to be "CR-10 Max".

Hope this helps, thanks to all the contributors who make Marlin Possible, YOU ROCK! :b

### Benefits

- Makes BLTouch Probe work on the CR-10 Max
- Makes the printer name "CR-10 Max"

### Related Issues

- None as far as I know

### DIFF
--- a/config/examples/Creality/CR-10 Max/Configuration.h
+++ b/config/examples/Creality/CR-10 Max/Configuration.h
@@ -127,7 +127,7 @@
 //#define BLUETOOTH
 
 // Name displayed in the LCD "Ready" message and Info menu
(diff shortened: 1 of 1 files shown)


ellensp:
Nice catch, thanks.

DakotaCodes00:
@ellensp No problem, glad I was able to help :)

PR #982, opened 2023-10-22
Remove Period

thisiskeithb:
### Description

Remove period from "Use compressed data to save Flash space" comment since it is not upstream: https://github.com/MarlinFirmware/Marlin/blob/b8b1aa345cfeaa59e9d377edc3c143fc67cf16cc/Marlin/Configuration.h#L3337

```c
  #define COMPACT_MARLIN_BOOT_LOGO  // Use compressed data to save Flash space
```

...or in the default config hosted here: https://github.com/MarlinFirmware/Configurations/blob/cfd3d8212aea29477c21fb0c71c886884071c8e8/config/default/Configuration.h#L3369

### Benefits

One fewer item to fix while comparing configs to upstream/main repo.

### DIFF
--- a/config/examples/3DFabXYZ/Migbot/Configuration.h
+++ b/config/examples/3DFabXYZ/Migbot/Configuration.h
@@ -3399,7 +3399,7 @@
 
   //#define TFT_SHARED_IO   // I/O is shared between TFT display and other devices. Disable async data transfer.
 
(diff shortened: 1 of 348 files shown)


thisiskeithb:
Oh boy. That's going to try and build every single config. 🙃

ellensp:
Nice trick to find the broken configs though :)

PR #985, opened 2023-10-24
Enable BLTouch 5V Mode CR-10 Max

DakotaCodes00:
### Description

Apologies for making multiple pull requests, I am finding these little things as I'm tinkering with my CR-10 Max :b

Enabled BLTouch 5V Mode as the CR-10 Max Mainboard (Creality V2.4) supports it by default, without this change the BLTouch works for the most part but ignores the Y offset of the NOZZLE_TO_PROBE_OFFSET option found in the Configuration.h file from my experience.

I contacted Creality for other reasons but ended up getting the CR-10 Max Pinout from them and it explicitly labels the BLTouch voltage pin (red) as being 5V.

Here is the [picture](https://i.imgur.com/Zt0QaZC.png) for reference the labels are in Mandarin so google translate on mobile is a good tool to have :)

### Benefits

- Allows BLTouch to work as expected

### Related Issues

- N/A as far as I am aware

### DIFF
--- a/config/examples/Creality/CR-10 Max/Configuration_adv.h
+++ b/config/examples/Creality/CR-10 Max/Configuration_adv.h
@@ -982,7 +982,7 @@
    * differs, a mode set EEPROM write will be completed at initialization.
    * Use the option below to force an EEPROM write to a V3.1 probe regardless.
    */
-  //#define BLTOUCH_SET_5V_MODE
+  #define BLTOUCH_SET_5V_MODE
 
   // Safety: Enable voltage mode settings in the LCD menu.
   //#define BLTOUCH_LCD_VOLTAGE_MENU


ellensp:
The bltouch 5 volt mode is to set the output voltage on the bltouch trigger signal.  It has nothing to do with any offsets

It is also unrelated to the bltouch supply voltage.  They are all powered by 5v even on 3.3v systems. 


English version, and way better documented


creality_v24_board_pinout-11-1024x768

Because this is a mega2560, and all IO pins are 5v, enabling 5V Mode will not harm anything.

ellensp:
Your offset problem is way more likely that your updating Configuration.h offsets, but these offsets are in eeprom, so you need to refresh your eeprom every time you update them. Either manually M502 then M500 or using  EEPROM_INIT_NOW

ellensp:
From the Bltouch site

```
Open Drain for Logic Voltage Free

​

In the BLTouch, an additional MOSFET is used to implement the Logic Free 3.3V / 5V (Open Drain) output mode.

​

Open Drain(Default) mode depends on the logic level of your control board, so if your board is a 5V logic system, BLTouch operates in 5V logic mode and 3.3V logic mode if your board is a 3.3V logic system.


Other than some special cases  (for example, a controller board that has a Zmin input with a large capacitance), the Smart V3 is a drop-in replacement for a V2 (no firmware modification needed).
```

There are no capacitors on the trigger IO line, this is not required.

DakotaCodes00:
Hi @ellensp thanks for taking the time to review and reply to my PR I appreciate it.

I did do a manual EEPROM initialization after every flash and I do have the `EEPROM_INIT_NOW` option enabled anyway, the issue only resolved itself after configuring the 5V option for the BLTouch somehow. Not sure why or how but it did in my case. 

Thanks for the better pinout for the Creality V2.4 main board helps a ton! 

If this 5V option is not a necessary change I am happy to close the PR :b

Apologies for any misunderstandings, hope you are having a good day!

PR #987, opened 2023-11-06
Fix TMC2225 *_RSENSE Values for BTT/Biqu Configs

thisiskeithb:
### Description

While TMC2225s are configured like TMC2208s, BTT's TMC2225s use 0.15 RSENSE resistors instead of 0.11.

### Benefits

Biqu/BTT configs with TMC2225s will have proper RSENSE values.

### DIFF
--- a/config/examples/BIQU/B1 SE Plus/Configuration_adv.h
+++ b/config/examples/BIQU/B1 SE Plus/Configuration_adv.h
@@ -2925,7 +2925,7 @@
     #define X_CURRENT       800        // (mA) RMS current. Multiply by 1.414 for peak current.
     #define X_CURRENT_HOME  X_CURRENT  // (mA) RMS current for sensorless homing
     #define X_MICROSTEPS     16        // 0..256
-    #define X_RSENSE          0.11     // Multiplied x1000 for TMC26X
+    #define X_RSENSE          0.15     // Multiplied x1000 for TMC26X
     #define X_CHAIN_POS      -1        // -1..0: Not chained. 1: MCU MOSI connected. 2: Next in chain, ...
     //#define X_INTERPOLATE  true      // Enable to override 'INTERPOLATE' for the X axis
     //#define X_HOLD_MULTIPLIER 0.5    // Enable to override 'HOLD_MULTIPLIER' for the X axis
@@ -2945,7 +2945,7 @@
     #define Y_CURRENT       800
     #define Y_CURRENT_HOME  Y_CURRENT
     #define Y_MICROSTEPS     16
-    #define Y_RSENSE          0.11
+    #define Y_RSENSE          0.15
     #define Y_CHAIN_POS      -1
     //#define Y_INTERPOLATE  true
     //#define Y_HOLD_MULTIPLIER 0.5
@@ -2965,7 +2965,7 @@
     #define Z_CURRENT       800
     #define Z_CURRENT_HOME  Z_CURRENT
     #define Z_MICROSTEPS     16
-    #define Z_RSENSE          0.11
+    #define Z_RSENSE          0.15
     #define Z_CHAIN_POS      -1
     //#define Z_INTERPOLATE  true
     //#define Z_HOLD_MULTIPLIER 0.5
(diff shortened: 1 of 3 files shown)


vovodroid:
If this value is specific for BTT's TMC2225s, may be it's worth to explicitly create configuration for this driver?
Because it's always possible to replace BTT2205 driver with anything else and thus again wrong value will be used.

thisiskeithb:
> Because it's always possible to replace BTT2205 driver with anything else and thus again wrong value will be used.

If you replace this kind of hardware, you need to update the config, so that is assumed.

vovodroid:
> If you replace this kind of hardware, you need to update the config, so that is assumed.

Sure, but it's better to do this with some define, but not with manual edit of specific parameter.

thisiskeithb:
Sure, but it's better to do this with some define, but not with manual edit of specific parameter.

`#define *_RSENSE` *is* the define.

You have to change if you're using anything outside of the usual 0.11 value like with these TMC2225s, 5160s, some kind of specific board design like the Einsy Rambo with 0.22 resistors, etc.

vovodroid:
Well, I don't see any usage if this value anywhere in Marlin code besides config_adv.h anyway... https://github.com/search?q=repo%3AMarlinFirmware%2FMarlin+X_RSENSE&type=code

Do I miss something?

GMagician:
Search for ST##_RSENSE.
'##' is for concatenation ST is define argument

vovodroid:
So changing rsence define from 0.11(+0.02) to 0.15(+0.02) increases Vref  (and current) for about 30%, doesn't it?

GMagician:
Value in define or real resistor? Because rsense value should be the resistor mounted on stepstick. If you want to increase current you have to change X_CURRENT value

vovodroid:
Define only. Resistor is the same, but define is changed from 0.11 to 0.15. According to formula it increases Vref, and thus current. Do I understand it correctly?

GMagician:
?_rsense is used by code to correctly convert current you define in ?_CURRENT into proper tmc register value. Why not simply increase ?_current value?

vovodroid:
> Why not simply increase ?_current value?

That's exactly what I'm talking about - I want to keep current current ))), even it's not correct,  but this fix will increase it by factor 1.3, as far as I understand. Should I decrease all current values by 1.3?

GMagician:
Ok, I just walked up all the messages, if your stepstick mounts a 0.15 resistor changing define value should fix and when you set current to a value it should give such current. I've not read 2225 datasheet to see if computation is correct so maybe fixing rsense  may not solve the problem at all but looking at btt github they post an example and say that if you set 800mA you get 800mA (and if picture you can see 0.15 on rsense)

immagine

thisiskeithb:
>  I want to keep current current ))), even it's not correct

How hot do your motors get after correcting the RSENSE values?

My B1 SE (beta unit) has an SKR 1.4 with TMC2208s & A4988s and my B1 SE Plus has TMC2209s in it now, so I can't do a 1:1 check at the moment.

At any rate, you'd decrease X/Y/Z's 800 mA to 586 mA and Z's 950 mA to 696 mA if you want the "old" final run current.

vovodroid:
> How hot do your motors get after correcting the RSENSE values?

I still didn't flash and test. But from 

Untitled

and thus Vref = Irms*(Rsense+0.02)*some_coefficient

I conclude that increasing R_SENSE from 0.11 to 0.15 increases Vref and current by 1.3. Correct? So now current will be according to defined, but in the same time it will suddenly increase comparing to current before fix.

thisiskeithb:
> I still didn't flash and test.

Oh. I assumed they were overheating based on your concerns.

> I conclude that increasing R_SENSE from 0.11 to 0.15 increases Vref and current by 1.3. Correct?

Yes. It's ~30% increase.

> So now current will be according to defined, but in the same time it will suddenly increase comparing to current before fix.

Correct & that is expected. They were running ~30% under expected current before.

vovodroid:
>They were running ~30% under expected current before
And it was more than enough. No skipping, and I even reduced Y and E current to keep motors cooler (though they were about 50 deg.C, less than max temperature).

PR #995, opened 2023-12-24
add config for Creality E3 free-runs board

mcrossen:
A few days ago, support for the [Creality "E3 free-runs TMC2209"](https://store.creality.com/products/e3-free-runs-tmc2209-32-bit-open-source-silent-motherboard) was added to marlin firmware ([PR link](https://github.com/MarlinFirmware/Marlin/pull/25636)) with a bug fix on the way ([PR link](https://github.com/MarlinFirmware/Marlin/pull/26562)). Now that support is added, the configs should be similar to other ender 3 configs so I started by copying the configs from the CrealityV427 board.

The main difference between configs for this new board and other Ender 3 boards are the following
`#define MOTHERBOARD BOARD_CREALITY_CR4NTXXC10`
`#define X_DRIVER_TYPE  TMC2209`
`#define Y_DRIVER_TYPE  TMC2209`
`#define Z_DRIVER_TYPE  TMC2209`
`#define E0_DRIVER_TYPE TMC2209`

Creality's official [repo](https://github.com/CrealityOfficial/E3-Free-runs-Silent-Motherboard/) for the board has an outdated marlin version, but it should still be a good reference to check other config values. I `diff`ed the config in that repo with a newer marlin version to try and find any other config values that should be changed. There was a lot of non-functional changes (like changes to comments and changes to standard config) to sort out but the following seemed relevant so i added it to this PR.
`#define X_CURRENT       800`
`#define Y_CURRENT       800`
`#define Z_CURRENT       800`
`#define  X_SLAVE_ADDRESS 0`
`#define  Y_SLAVE_ADDRESS 1`
`#define  Z_SLAVE_ADDRESS 2`
`#define E0_SLAVE_ADDRESS 3`

Perhaps someone more familiar with these values can verify that they're necessary. I also found the following differences in Creality's official repo but decided NOT to add them because they didn't seem relevant.
`#define BABYSTEPPING` // disabled in Creality's repo but not for CrealityV427

I tested this code on my ender 3 pro by applying a few configs specific to my printer (like bltouch), flashing the firmware, and printing a benchy. I'm excited to report that the new [silent] board can't be heard throughout the whole apartment :smile:


ellensp:
The _CURRENT is  dependent on the stepper motors themselves and their peak current settings.   Not a independent stepper driver setting

ellensp:
One thing you do need to change is  
#define X_RSENSE          0.11     // Multiplied x1000 for TMC26X

The circuit diagram shows this board has 0.15 ohm resistors    

And Y, Z and E

ellensp:
setting _SLAVE_ADDRESS In configuration_adv.h is not required, it is set automatically in Marlin/src/pins/stm32f4/pins_CREALITY_CR4NTXXC10.h  

ie 
```CPP
#ifndef X_SLAVE_ADDRESS
    #define X_SLAVE_ADDRESS                   0
  #endif
  #ifndef Y_SLAVE_ADDRESS
    #define Y_SLAVE_ADDRESS                   1
  #endif
  #ifndef Z_SLAVE_ADDRESS
    #define Z_SLAVE_ADDRESS                   2
  #endif
  #ifndef E0_SLAVE_ADDRESS
    #define E0_SLAVE_ADDRESS                  3
  #endif

```

mcrossen:
alright, thanks @ellensp. I modified my commits with those changes. I might not have understood what you meant for X_RSENSE. can you verify I changed it correctly?

ellensp:
Yes that is good.

mcrossen:
One other change I forgot to mention is that Creality's repo disabled software fan pulse-width-modulation. Most ender 3 boards have it enabled but this Pull Request disables it to match Creality.

`//#define FAN_SOFT_PWM`

mcrossen:
here's the results of my tests. The left benchy was printed at the start of this PR. The right benchy includes all recent changes. Notice how the left benchy has a horizontal line on the hull. That is from the power supply cutting out. Fortunately, I have power supply recovery enabled which is why it was able to finish. The right benchy printed without any power issues - likely thanks to @ellensp's suggestions about XYZE_CURRENT and XYZE_RSENSE.

ellensp:
you probably want to enable the following, as TMC2209s like this mode.  (and halves the stepping load on marlin) 

```
  /**
   * Step on both rising and falling edge signals (as with a square wave).
   */
  #define EDGE_STEPPING
```
And the following, to reduce controller CPU load 

```
#define ENDSTOP_INTERRUPTS_FEATURE
```

mcrossen:
I sent @jghboke a config to try on their ender 5 with this board. They say it works so i've added it to this PR.

mcrossen:
can anyone review this PR?

PR #997, opened 2023-12-26
Sovol SV-05

MRR0GERS:
### Description

Added new stock configs for Sovol SV-05.  The SV-05 is similar to the Creality Ender 5 Pro but when I used those configurations and compiled a new firmware the SV-05 would not work correctly:
  - would not level with probe
  - bed would not raise and would go down with each attempt at auto home

I used a clean copy of the default config and copied properties from Sovol's SV--05 configuration build on their support page.

### Benefits

SV-05 worked correctly after new firmware flash.

### Related Issues

Latest Marlin build with stock settings for Sovol SV-05 using the Creality Marlin 4.2.2 mainboard (comes standard)


thinkyhead:
Does this mean the `CREALITY_RTS` display option is actually working in our un-merged PR MarlinFirmware/Marlin#25382 ? It's been a while since I visited that PR, and at that time I was still cleaning up the code and looking for testers.

thinkyhead:
n/m - Commented on the wrong PR.

PR #1003, opened 2024-01-17
🩹 Fix Robo3D R1+ Config

thisiskeithb:
### Description

As discovered on Discord, this config had some minor issues like using `true`/`false` incorrectly that I've patched here. I also stripped out the `// ncrmro` comments since Nicholas' credit is already in the `STRING_CONFIG_H_AUTHOR` define.

### Benefits

Correct / clean config.

### DIFF
--- a/config/examples/Robo3D/R1+/Configuration.h
+++ b/config/examples/Robo3D/R1+/Configuration.h
@@ -563,7 +563,7 @@
 #define TEMP_SENSOR_5 0
 #define TEMP_SENSOR_6 0
 #define TEMP_SENSOR_7 0
-#define TEMP_SENSOR_BED 1 // ncrmro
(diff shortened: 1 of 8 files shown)


config/examples/Robo3D/R1+/Configuration.h
```diff
@@ -1859,7 +1859,7 @@
  */
 
 // Min software endstops constrain movement within minimum coordinate bounds
-#define MIN_SOFTWARE_ENDSTOPS false // ncrmro
+#define MIN_SOFTWARE_ENDSTOPS
```
thisiskeithb, line 1862: This was set to `false`, but I didn't disable it here since the machine homes to MIN and does not have extra/negative X/Y/Z travel.

PR #1012, opened 2024-01-28
Various config fixes

thisiskeithb:
### Description

Various config fixes after running the [`build_all_examples`](https://github.com/MarlinFirmware/Marlin/blob/bugfix-2.1.x/buildroot/bin/build_all_examples) script.

Changes should be obvious, but I dropped `PREHEAT_2_LABEL` / ABS items on a few configs since they were too low for ABS after making the max temp sanity check happy.

Remaining configs with build errors:

- **FoamCutter/generic/** &mdash; _Z_MIN_PIN, Z_STOP_PIN, or Z_SPI_SENSORLESS is required for Z axis homing_
- **FoamCutter/rcKeith/** &mdash; _Z_MIN_PIN, Z_STOP_PIN, or Z_SPI_SENSORLESS is required for Z axis homing_
- **Geeetech/A10M/YHCB2004_V4.1/** &mdash; _undefined reference to `operator delete(void*, unsigned int)'._
   See https://github.com/MarlinFirmware/Marlin/issues/26571
- **Geeetech/A10PRO/** &mdash; _undefined reference to `operator delete(void*, unsigned int)'._
   See https://github.com/MarlinFirmware/Marlin/issues/26571
- **Geeetech/A10T/YHCB2004_V4.1/** &mdash; _undefined reference to `operator delete(void*, unsigned int)'._
   See https://github.com/MarlinFirmware/Marlin/issues/26571

Configs with pending upstream (Marlin) support:

- **Creality/CR-6 SE/**
- **Creality/Ender-5 S1/**

Configs with non-standard warnings (not fixed here):

- **Creality/CR-10 Max/** &mdash; _comparison between signed and unsigned integer expressions_
- **Creality/CR-10S Pro/** &mdash; _comparison between signed and unsigned integer expressions_
- **Creality/CR-20/RepRapWorld Minitronics20/** &mdash; _"LITTLE_ENDIAN" redefined / deleting object of polymorphic class type 'DoubleBufferedEPOutHandler' which has non-virtual destructor might cause undefined behavior_
- **Creality/CR-30 PrintMill/** &mdash; _unused variable 'z_raised'_
- **Creality/Ender-5 Plus/CrealityV1 (RELOADED UI)/** &mdash; _comparison between signed and unsigned integer expressions_
- **MakerParts/** &mdash; _Various framework-arduino-sam warnings_
- **Malyan/M200/** &mdash; _No CMSIS startup file defined, custom one should be used_
[...]

PR #1024, opened 2024-02-21
🐛 Disable Z homing on foamcutters

DerAndere1:
### Description

Don't home dummy Z axis for foamcutters.

### Benefits

fix compilaton with foamcutter configs (bug found by @thisiskeithb )

### Related Issues

- https://github.com/MarlinFirmware/Configurations/pull/1012
- https://github.com/MarlinFirmware/Marlin/pull/24325#discussion_r1469935776


thisiskeithb:
Ahh, so these particular examples don't home Z. Some foam cutters *do* home Z, so that was confusing 😄

PR #1026, opened 2024-02-29
Creality Ender 3 Pro / CR-10 S5 / 427 + BLTouch

JPT77:
### Requirements
Creality printers with STM32 version 4.2.7 board and BL touch sensor.
### Description
This config adds a BL touch config to the above two creality printers . It is not 100% final but it's already quite stable. 
I added a lot of comments with "JPT" which explain why I set some variables this way and where more work my be needed. I also added a read-me and fetched the logo files from existing configs.
The files are meant for current 2.1-bugfix branch, not the 2.1.2.2 release. Should I deliver configs for the latest release too? Where?
### Benefits
Will be easier to set up BL touch for these printers. 
### Disclaimer
It's 10 years since I last fiddled with Marlin. So please check thoroughly before accepting this PR.


thinkyhead:
It is best that the configuration with BLTouch match as closely as possible the parent config in all aspects that are not probe-related. If there are improvements to be made to a given machine configuration, it should be made to all of the machine's configurations.

thinkyhead:
Is the bed temp sensor actually different on this model (1 instead of 5), with a higher max temp?

thinkyhead:
- When using UBL the mesh should occupy the entire bed area (minus a safe margin).
- As you noted, the MAX POS values should almost always be larger than the BED SIZE values. Those should be corrected in the standard configs also. I just did this for X MAX only so far.
- Although _your_ machine can't handle a high BAUD RATE we should leave it alone and stick with the defaults. The same goes for any other adjustments that are not universally required.

JPT77:
> Is the bed temp sensor actually different on this model (1 instead of 5), with a higher max temp?

I don't know. 
I tried checking the bed temp with an infrared thermometer and it seemed to be almost 10 °C off. Seemed to be less with the new config. but not sure.

> * When using UBL the mesh should occupy the entire bed area (minus a safe margin).

minus save margin or minus "reachable"?
ie, you you mean it should include positions not reachable by the probe that should be taken manually?

> * As you noted, the MAX POS values should almost always be larger than the BED SIZE values. Those should be corrected in the standard configs also. I just did this for X MAX only so far.

not sure what that is about, already forgot ;)
> * Although _your_ machine can't handle a high BAUD RATE we should leave it alone and stick with the defaults. The same goes for any other adjustments that are not universally required.

I agree. seems to be another issue on my part. 

did you change everything you mentioned here or should I?

thinkyhead:
> minus save margin or minus "reachable"?

The "safe margin" is a margin that ensures that the probe gets good readings when it probes the bed. Inductive probes don't do well near the edge, and some cheaper beds have more outlier heights nearer to the edge.

Which points are reachable will depend on where the probe is at the extremes. If the probe can't reach some edge of the bed, we don't care for UBL. With UBL the points the probe can't reach can be probed afterward with the nozzle paper test. For this reason the UBL mesh should cover almost the entire bed, minus that "safe margin" to ensure the probe doesn't freak out at the edges.

PR #1031, opened 2024-03-21
Add CR-10S with BTT SKR Mini E3 3.0, Ender3 V2 with BTT SKR Mini E3 2.0, Ender3 V2 with creality v4.2.7

ellensp:
### Requirements

Users requested these in "[FR] Configurations for all 3D printers and every possible permutation" 

CR-10S with BTT SKR Mini E3 3.0
Ender3 V2 with BTT SKR Mini E3 2.0
Ender3 V2 with creality v4.2.7

### Description

added CR-10S BTT SKR Mini E3 3.0 configs
addee E3V2 BTT SKR Mini E3 2.0 configs
added E3V2 creality v4.2.7 configs

### Benefits

More common printer and motherboard combinations

### Related Issues
https://github.com/MarlinFirmware/Configurations/issues/649

### NOTE
All untested beyond building. These where trivial changes from existing example configs


ellensp:
Error is deliberate to ensure user knows they need a custom display cable

config/examples/Creality/CR-10S/BigTreeTech SKR Mini E3 3.0/Configuration_adv.h
```diff
@@ -575,7 +575,7 @@
  */
 #define USE_CONTROLLER_FAN
 #if ENABLED(USE_CONTROLLER_FAN)
-  //#define CONTROLLER_FAN_PIN -1           // Set a custom pin for the controller fan
+  #define CONTROLLER_FAN_PIN FAN1_PIN       // Set a custom pin for the controller fan
```
thisiskeithb: For the SKR Mini E3 V3-based configs, you’d typically set this to `FAN2_PIN` and use `FAN1_PIN` for `E0_AUTO_FAN_PIN`.

ellensp: Yes I see... though there is of course one eg that doesn't (ender 3 v2)

thisiskeithb:
> Error is deliberate to ensure user knows they need a custom display cable

The CI test could be updated here to skip fatal warnings like we do for the [`build_example` script](https://github.com/MarlinFirmware/Marlin/blob/05896551f1fd521865e6cc056a3acc07a6bf2c0b/buildroot/bin/build_example#L37-L38):

```bash
# Suppress fatal warnings
echo -e "\n#define NO_CONTROLLER_CUSTOM_WIRING_WARNING" >> Marlin/Configuration.h
```

This is especially true for future ambitions to automatically build configs in regular intervals and eventually automate config releases when there’s a new tag upstream.

Also, adding the disabled defines in 8b9e592 doesn’t seem necessary.

ellensp:
"Also, adding the disabled defines in https://github.com/MarlinFirmware/Configurations/commit/8b9e59247bb8c6d73f790971524a82e800b6e221 doesn’t seem necessary."

Its to stop uses going I cant find it..   now its right there in their faces.

thisiskeithb:
> Its to stop uses going I cant find it.. now its right there in their faces.

Now we have another place to remember to update if the name changes since it’s not included as a config item / in all configs by default.

PR #1044, opened 2024-04-21
Add Configuration files for Two Trees SP-5 V3

ellensp:
### Requirements

Two Trees SP-5 V3

### Description

As requested, here are the Two Trees SP-5 V3 Configuration files

### Benefits

We have these Config files

### Related Issues
https://github.com/MarlinFirmware/Configurations/issues/649


thinkyhead:
I've added a note to the README that 🎵  someday 🎵 we might fully support that UI (after there's a "final" revision of the UI?).

ellensp:
[SP-5-nano-v1.2-3x22082x2225Firmware.zip](https://github.com/user-attachments/files/16321316/SP-5-nano-v1.2-3x22082x2225Firmware.zip)

ellensp:
Two Trees SP-5 V3 source [Mks-Robin-Nano-Marlin2.0-Firmware-master.zip](https://github.com/user-attachments/files/16321445/Mks-Robin-Nano-Marlin2.0-Firmware-master.zip)
Rename to zipx (required zipx to get the size under github file limit)

config/examples/Two Trees/SP-5_V3/README.md
Alexander74, line 1: Hi. Link to "Original Marlin code from Two Trees" is expired.
Can you provide actual link or config files?

ellensp: It is attached to the github right above this post, in zipx format (as that was the only way to get it small enough to attach to github)

PR #1045, opened 2024-04-24
🔧 Suppress BIQU Hurakan MicroProbe pull-up warning

thisiskeithb:
### Description

This is a working OEM config and shipped with the MicroProbe connected to the PROBE port.

### Benefits

Prevents confusion since users compiling a stock/OEM Hurakan config do not need to see this warning.


thisiskeithb:
Related to this, I've been thinking about adding `DIAG_JUMPERS_REMOVED` / `DIAG_PINS_REMOVED` / `NO_AUTO_ASSIGN_WARNING` to all OEM configs as well since users don't need to know that Marlin is remapping things behind the scenes for an OEM/stock config and the printer ships in a usable state without needing to mod or change anything.

PR #1046, opened 2024-04-24
🔧 Fix broken Velleman K8400 links

albertskog:
### Requirements

### Description

Fix broken links.

### Benefits

Links now work.


thisiskeithb:
The `vertex-m1-v1.4-h2.zip` links under the "Original Sources" header appear to be dead as well.

Also, these READMEs are identical and could be unified & brought up one level so a single README covers both configs.

albertskog:
Oops, did not think to check the other links! The one at the top was dead as well. Fixed them all and moved up one level as suggested.

thisiskeithb:
> For implementation and updated K8400 firmware, see https://github.com/birkett/Vertex-K8400-Firmware

Doesn't look like this is updated anymore:

image

...which is OK since *this* repo (`MarlinFirmware/Configurations`) serves as the matched configs repository for upstream Marlin and is kept up to date 😄

albertskog:
Better like this?

PR #1047, opened 2024-04-25
🔧 FlashForge Creator Pro updates

queengooborg:
### Description

This PR updates the configuration data for the FlashForge Creator Pro, following issues I ran into during my own work modding my FFCP.  This performs the following updates:

- Updates the readme for better legibility and formatting
  - Also defines which extruder is which
- Renames the machine to "FF Creator Pro"
- Updates the author field to add my name
  - Originally, I planned to leave this as-is, but since I've made a fair number of tweaks to the configuration, I think it makes more sense to add my name!
- Enables speaker support, which the FFCP has, so that M300 can be used
- Increase Z-axis homing feedrate slightly
  - The original feedrate was a bit slow for my taste
- De-center origin and move origin to bottom-left corner
  - This one may seem like a bizarre change at first, however there's a few main reasons why I made this change
    - When doing a test print, I found that the original positioning values were not actually quite centered, so anything that was printed on the right edge may print off the build volume
    - The bottom-left corner is a far easier point to measure than the center, so this allows for easier maintenance down the line
    - Most other 3D printers appear to also have their origin in the bottom-left corner now, as far as I can tell
- Handle nozzle offset in firmware
  - The current guide instructs the user to set the nozzle offset in their slicer; however, Ultimaker Cura shrinks the build volume if there is a nozzle offset, as it assumes that neither nozzle can go outside of the build volume
  - The original author appears to account for this by informing the user to add double the offset to their build volume, but if only one nozzle is used, only 1x offset is snipped off the build volume, leading to the potential of placing objects outside of the actual build volume
  - Accounting for the offset in firmware will eliminate this issue altogether
  - The original FlashForge firmware also performs a firmware nozzle offset, so this mimics the original firmware better
- Adds some GCode to produce tones on job start, completion and cancellation
  - The job start tone is based on the original FlashForge Creator Pro jingle
- Adds bed leveling and tramming support

### Benefits

Benefits stated in description!


queengooborg:
Since changes are combined in this PR now, let me make a few more nice-to-have tweaks real quick!

queengooborg:
All set now!

thisiskeithb:
> Fixes the left extruder's direction (I found that it is reversed)

Were both extruder directions incorrect? I'm curious why no one has run into this issue before. Also, you updated E0 and E2 (I think you meant to update E1 here).

queengooborg:
Just the one extruder motor -- I'm kind of surprised that no one else caught this, since the extruder motors need to spin in opposite directions from each other due to the way they grab the filament.  Maybe I'm the first to do a Marlin mod and try to use both extruders!  😆 

Its also very possible that the connector to the stepper driver was flipped by the manufacturer, and when I put my own printer back together, I un-flipped it.  Happy to revert if so!

---

Regarding E0, that was accidental (I flipped that boolean in my own configuration while working on a mod to make my left extruder the first extruder), but it doesn't seem to have any effect.  Happy to revert that as well!

thisiskeithb:
There is no E2, only E0 and E1 (zero-based), so changing `INVERT_E2_DIR` doesn't actually do anything.

> Regarding E0, that was accidental [...], but it doesn't seem to have any effect

Please push another commit that fixes the correct extruder behavior & reverts the change to the unaffected extruder.

queengooborg:
It actually _did_ make a difference on my setup, even before applying left-extruder-is-first modding.  I'm a bit too far into that to go back and test without it, so I've just reverted the changes!

thisiskeithb:
> It actually did make a difference on my setup

That wouldn't be possible unless you're running three extruders: E0, E1, and E2.

Also, from https://github.com/MarlinFirmware/Marlin/pull/12855:

> Dual extrusion worked right on Simplify and the CreatorPro

So maybe there is something different about your printer?

queengooborg:
Unless I somehow got a non-standard unit from Amazon, there shouldn't have been anything different!

Maybe I changed some other variable besides the `E2` variable and hadn't realized it?

Edit: okay, I just looked at my change history, and yes, it looks like I did change `E0` when doing those tests and didn't realize it.  Whoops!

queengooborg:
I've pushed a bunch more updates to get the FFCP configuration working as desirably as possible -- some of them may be more of a personal preference than others (ex. the job start chime), but I feel all the changes are helpful!

queengooborg:
Thanks for merging!  I feel embarrassed with how many typos there were in the documentation, haha...!

PR #1048, opened 2024-04-26
FlashForge Creator Pro: has a speaker

queengooborg:
### Description

The FlashForge Creator Pro has a speaker that can produce tones, not just a one-tone buzzer.  This updates the configuration accordingly.

### Benefits

Naturally, this allows the use of M300 as well as for the Marlin firmware to produce additional tones.


thisiskeithb:
Combined into https://github.com/MarlinFirmware/Configurations/pull/1047

Issue #1052, opened 2024-05-09
Build All Configurations Status - May 2024


thisiskeithb:
## Description

Results of Marlin's [`build_all_examples`](https://github.com/MarlinFirmware/Marlin/blob/bugfix-2.1.x/buildroot/bin/build_all_examples) script.

### Creality/CR-6 SE

> [!NOTE]  
> Needs https://github.com/MarlinFirmware/Marlin/pull/19958

### Creality/Ender-5 S1

> [!NOTE]  
> Needs https://github.com/MarlinFirmware/Marlin/pull/25382

### Geeetech/A10M/YHCB2004_V4.1

> [!NOTE]  
> See https://github.com/MarlinFirmware/Marlin/issues/26571#issuecomment-2101314494 for open issue & possible fix.

<details><summary>details:</summary>
<p>

```prolog
/var/folders/l2/mtr_0zp945s_9s6s77z1d2fw0000gn/T//ccpC0Ipr.ltrans2.ltrans.o: In function `global destructors keyed to 65535_1_HAL.cpp.o.18264':
<artificial>:(.text.exit+0x28): undefined reference to `operator delete(void*, unsigned int)'
collect2: error: ld returned 1 exit status
*** [.pio/build/mega2560/firmware.elf] Error 1
============================================================== [FAILED] Took 25.34 seconds ==============================================================

Environment    Status    Duration
-------------  --------  ------------
mega2560       FAILED    00:00:25.335
========================================================= 1 failed, 0 succeeded in 00:00:25.335 =========================================================
Failed
```

</p>
</details>

### Geeetech/A10PRO

> [!NOTE]  
> See https://github.com/MarlinFirmware/Marlin/issues/26571#issuecomment-2101314494 for open issue & possible fix.

<details><summary>details:</summary>
<p>

```prolog
/var/folders/l2/mtr_0zp945s_9s6s77z1d2fw0000gn/T//ccxzR0nA.ltrans2.ltrans.o: In function `global destructors keyed to 65535_1_HAL.cpp.o.17263':
<artificial>:(.text.exit+0x20): undefined reference to `operator delete(void*, unsigned int)'
collect2: error: ld returned 1 exit status
*** [.pio/build/mega2560/firmware.elf] Error 1
============================================================== [FAILED] Took 18.53 seconds ==============================================================

Environment    Status    Duration
-------------  --------  ------------
mega2560       FAILED    00:00:18.535
[...]

PR #1056, opened 2024-05-21
🩹 Fix Formbot T-Rex 2+ X2 max position

thisiskeithb:
### Description

While running the [`build_all_examples`](https://github.com/MarlinFirmware/Marlin/blob/bugfix-2.1.x/buildroot/bin/build_all_examples) script, this config would not build:

```prolog
In file included from Marlin/src/HAL/AVR/../../inc/MarlinConfigPre.h:56:0,
                 from Marlin/src/HAL/AVR/../../inc/MarlinConfig.h:28,
                 from Marlin/src/HAL/AVR/HAL.cpp:24:
Marlin/src/HAL/AVR/../../inc/../../Configuration_adv.h:823:24: error: floating constant in preprocessor expression
   #define X2_MAX_POS   434.6      // The max position of the X2 carriage, typically also the home position
                        ^
```

Rounding `X2_MAX_POS` down to `434` prevents the error and allows the config to build.

### Benefits

Config now builds.

### Related Issues

- #1052

### DIFF
--- a/config/examples/Formbot/T_Rex_2+/Configuration_adv.h
+++ b/config/examples/Formbot/T_Rex_2+/Configuration_adv.h
@@ -820,7 +820,7 @@
   #define X1_MIN_POS X_MIN_POS    // Set to X_MIN_POS
   #define X1_MAX_POS (X_BED_SIZE) // A max coordinate so the X1 carriage can't hit the parked X2 carriage
   #define X2_MIN_POS     0        // A min coordinate so the X2 carriage can't hit the parked X1 carriage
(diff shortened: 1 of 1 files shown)


thisiskeithb:
Talking with @InsanityAutomation, he says that we need to allow floats in MIN/MAX positions for proper IDEX operation, so that will need to get fixed upstream.

thisiskeithb:
Apparently these TREX configs are *very* broken according to @InsanityAutomation, so I'll just close this for now as they won't work anyway.

PR #1063, opened 2024-06-03
Update AnyCubic Vyper configuration

ngraziano:
### Description

Configuration update for AnyCubic Vyper

### Benefits

 - Activate cooling fan for extruder 0, with actual config the fan never start (I don't understand why this is commented)
 - Activate tare with value from OEM firmware, it is needed for the probing to complete. If not set the filament pull on the head and the probe fail when the head is in the middle.
 - Disable leveling fade by default (like OEM firmware), because the the probed position is more than 2 mm bellow the zero point and using the printer just after G29 cause massive z error with actual value of 10

### Related Issues

None


thinkyhead:
Thanks for the patches! This will be a great boon to users of Marlin on Vyper. Maybe that E0 auto fan pin was commented because the original author wasn't confident….? Anyway, good to have it working properly!

PR #1070, opened 2024-06-25
🎨 Use standard STRING_CONFIG_H_AUTHOR format

thisiskeithb:
### Description

Wrap `STRING_CONFIG_H_AUTHOR` string in parenthesis per default & most configs.

### Benefits

Configs will match the standard `STRING_CONFIG_H_AUTHOR` format.

> [!NOTE]  
> Several configs will not build & will cause CI to fail, but they are a known issue. See https://github.com/MarlinFirmware/Configurations/issues/1068 for more details.

### DIFF
--- a/config/default/Configuration.h
+++ b/config/default/Configuration.h
@@ -63,7 +63,7 @@
 // @section info
 
 // Author info of this build printed to the host during boot and M115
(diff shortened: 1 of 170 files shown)


ellensp:
@thinkyhead If I wanted my name in the config I would have added it. I have removed them. 
I consider this a privacy violation.

PR #1074, opened 2024-06-27
🔧 Define Malyan M300 E0_AUTO_FAN_PIN directly

thisiskeithb:
### Description

Drop the alias for now in preparation for https://github.com/MarlinFirmware/Marlin/pull/27214

### Benefits

Upstream CI will pass 😄 

### Related Issues

- https://github.com/MarlinFirmware/Marlin/pull/27214

### DIFF
--- a/config/examples/delta/Malyan M300/Configuration_adv.h
+++ b/config/examples/delta/Malyan M300/Configuration_adv.h
@@ -681,7 +681,7 @@
  * Multiple extruders can be assigned to the same pin in which case
  * the fan will turn on when any selected extruder is above the threshold.
  */
-#define E0_AUTO_FAN_PIN AUTO_FAN_PIN
+#define E0_AUTO_FAN_PIN PA8
 #define E1_AUTO_FAN_PIN -1
 #define E2_AUTO_FAN_PIN -1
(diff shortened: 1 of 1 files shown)


thinkyhead:
The board provides a pin for this, and _we want boards to be able to provide these pins and have them automatically assigned_ without users needing to know the pin number. So the correct solution (for now) is to leave the pin undefined and let the pins file define it for us.

thisiskeithb:
> The board provides a pin for this, and _we want boards to be able to provide these pins and have them automatically assigned_ without users needing to know the pin number. So the correct solution (for now) is to leave the pin undefined and let the pins file define it for us.

The pins file [provides `AUTO_FAN_PIN`](https://github.com/MarlinFirmware/Marlin/blob/9240ec89d034632443d02baaeb140acc3134c67f/Marlin/src/pins/stm32f0/pins_MALYAN_M300.h#L93) which is not `E0_AUTO_FAN_PIN` and now this config is unsafe because it no longer enables the hotend cooling fan. I fixed this in https://github.com/MarlinFirmware/Marlin/pull/27214, which is why I was defining the pin directly here to ensure that the hotend auto fan still works for this config while things are hashed out in the linked PR.

PR #1078, opened 2024-07-12
🔥 Remove Formbot T-Rex configs

thisiskeithb:
### Description

Chatting with @InsanityAutomation, these configs do not match actual hardware, so let's remove them.

If/when someone contributes working configs that match actual hardware, we can bring back support for these machines.

### Benefits

Removes bad/invalid configs

### Related Issues

- #1068


thisiskeithb:
> This branch has conflicts that must be resolved

I'm not going to keep rebasing these configs after each change to `import-2.1.x` since these configs need to be completely removed. 😄

PR #1091, opened 2024-08-26
Sovol SV06 non-inverted logo

marbocub:
### Description

The logo of Sovol SV06 config is inverted:
image

It is non-inverted logo:
image

I think non-inverted is normal. This PR changes inverted logo to non-inverted one.


ellensp:
Yes this shouldn't be inverted like that, PR is correct.

PR #1110, opened 2024-10-09
Add TFT_SHARED_IO to Kingroon KP3S Example

carloslockward:
### Description

The Kingroon KP3S and other MKS ROBIN NANO based printers have been having issues with SD card prints for a couple of versions of Marlin. I have been trying different configs to see if the issue could be solved. After a lot of attempts I discovered that `TFT_SHARED_IO` seems to fix the issues. It would seem that the TFT and the SD card are sharing I/O. Therefore, this should be enabled in the example config.

We should also consider adding this to other examples of MKS ROBIN NANO based printers. But it should probably be on a case by case basis.

### Benefits

Without this config enabled the printer would refuse to print(go back to info screen and do nothing), freeze right before printing or freeze mid print(depending on the  Marlin version used)

### Related Issues

[[BUG] SD Card is detected but doesn't start printing or heating](https://github.com/MarlinFirmware/Marlin/issues/26913)
[[BUG] On Print Start from SD Card Print Never Starts](https://github.com/MarlinFirmware/Marlin/issues/26231)

### DIFF
--- a/config/examples/Kingroon/KP3S/Configuration.h
+++ b/config/examples/Kingroon/KP3S/Configuration.h
@@ -63,7 +63,7 @@
 // @section info
 
 // Author info of this build printed to the host during boot and M115
-#define STRING_CONFIG_H_AUTHOR "(carloslockward)" // Who made the changes.
+#define STRING_CONFIG_H_AUTHOR "(@AR1972)" // Who made the changes.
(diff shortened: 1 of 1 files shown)


ErikStroman:
Thank you @carloslockward! Saved me a long headache trying to sort through settings for my build with 2.1.3-b2. This is still a problem for the provided default config files for the KP3S Pro as of this beta build.

PR #1117, opened 2024-11-02
Set missing Longer LK & U20 SDIO define

tpruvot:
In the past (pre 2.1.2.1) we had a define SDIO_SUPPORT which was not replaced by ONBOARD_SDIO in the config samples nor the pins.h

so #define SDCARD_CONNECTION ONBOARD which do the job

### DIFF
--- a/config/examples/Alfawise/U20-bltouch/Configuration_adv.h
+++ b/config/examples/Alfawise/U20-bltouch/Configuration_adv.h
@@ -1948,7 +1948,7 @@
    *
    * :[ 'LCD', 'ONBOARD', 'CUSTOM_CABLE' ]
    */
-  //#define SDCARD_CONNECTION LCD
+  #define SDCARD_CONNECTION ONBOARD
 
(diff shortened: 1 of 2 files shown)


thinkyhead:
It looks like the only options would be `ONBOARD` and `CUSTOM_CABLE` since there's no standard SPI LCD/SD wiring specified. The board may be used in any machine, so `ONBOARD` is a sensible default when `SDSUPPORT` is enabled. So, should we (also) add this to `pins_LONGER_LK.h`? :
```cpp
#ifndef SDCARD_CONNECTION
  #define SDCARD_CONNECTION              ONBOARD
#endif
```

tpruvot:
The only SPI thing on this board is the eeprom, used to store some icons for the original UI, not at all in Marlin. The only way to use SPI is to unsolder the eeprom SMD chip so i guess we can even force ONBOARD_SDIO in pins_LONGER_LK.h for this one.

tpruvot:
To note : this board is one of the few ones with dedicated TFT pins only present on F103V and F103Z chips. Its not a SPI LCD

PR #1120, opened 2024-11-25
✨ Minitronics v1.1

lkundrak:
This basically matches what was originally shipped with the board back in the day.

I think it's nice to have it around. If anything perhaps it will make it easier to avoid regressing the board build.


thinkyhead:
Unable to edit the usual way, so ended up merging directly. Thanks!

PR #1125, opened 2025-01-26
Added CR-10 V2 config

callmeraymon:
### Description

Added configuration for Creality CR-10 V2 with BigTreeTech-SKR-Mini-E3-V3 board and BLTouch


thisiskeithb:
Thanks for your contribution! Unfortunately we can't accept PRs directed at release branches. We make patches to the import branches and only later do we push them out as releases. Please redo this PR starting with the `import-2.1.x` branch and be careful to target `import-2.1.x` when resubmitting the PR.

PR #1137, opened 2025-04-02
Fix unrealistic feedrates

AndreySamokhin:
### Description

[Adding a few additional sanity checks](https://github.com/MarlinFirmware/Marlin/pull/27764) for `MAXIMUM_STEPPER_RATE` revealed that some configuration files contain unrealistic feedrates. Without fixing these configuration files, the automatic checks cannot pass.

### Benefits

This change allows additional sanity checks to be added to the Marlin firmware, helping prevent unpredictable behavior by triggering a compile-time error when necessary.

### Related Issues

https://github.com/MarlinFirmware/Marlin/issues/27496
https://github.com/MarlinFirmware/Marlin/pull/27764


thinkyhead:
As a general note, very large feedrates on the E axis are expected and are meant to remove the E axis as a limitation on the other axis feedrates. So we should probably not make large E axis max feedrates into an error, and we can leave the very large E axis feedrates untouched.

AndreySamokhin:
I believe that even the adjusted values are still unrealistic.
Let's assume we are printing at $`200 \, \mathrm{mm/s}`$ with a $`1 \, \mathrm{mm}`$ nozzle (line width is $`1 \, \mathrm{mm}`$ and layer height is $`0.5 \, \mathrm{mm}`$). This results in a volumetric feedrate of $`100 \, \mathrm{mm^3/s}`$. In practice, we rarely exceed $`10 \, \mathrm{mm^3/s}`$.

In all the modified files, the specified feedrates are far beyond $`100 \, \mathrm{mm^3/s}`$:

*Creality/Ender-5 Plus/BigTreeTech SKR 3*, *Creality/Ender-5 Plus/CrealityV1 (ORIGIN UI)*, and *Creality/Ender-5 Plus/CrealityV1 (RELOADED UI)*:
$` \pi * (1.75 / 2)^2 * 5375 = 12928 \, \mathrm{mm^3/s}`$

*Velleman/K8400/Dual-head* and *Velleman/K8400/Single-head*:
$` \pi * (1.75 / 2)^2 * 1680 = 4041 \, \mathrm{mm^3/s}`$

*Wanhao/Duplicator i3 2.1*:
$` \pi * (1.75 / 2)^2 * 5000 = 12026 \, \mathrm{mm^3/s}`$

So, to me, it seems reasonable to reduce these values somewhat. That said, I am not insisting the proposed changes must be accepted. The same goes for the recently added `SANITY_CHECK_E_STEPPER_RATES` macro in *SanityCheck.h* (i.e., `#if ALL(HAS_EXTRUDERS, SANITY_CHECK_E_STEPPER_RATES)`). To me, `SANITY_CHECK_E_STEPPER_RATES` feels a bit overcautious. Anyway, should `//#define SANITY_CHECK_E_STEPPER_RATES` be added to *Configuration.h* in this case?

robherc:
> I believe that even the adjusted values are still unrealistic. Let's assume we are printing at 200 mm / s with a 1 mm nozzle (line width is 1 mm and layer height is 0.5 mm ). This results in a volumetric feedrate of 100 mm 3 / s . In practice, we rarely exceed 10 mm 3 / s .
> 
> In all the modified files, the specified feedrates are far beyond 100 mm 3 / s :
> 
> _Creality/Ender-5 Plus/BigTreeTech SKR 3_, _Creality/Ender-5 Plus/CrealityV1 (ORIGIN UI)_, and _Creality/Ender-5 Plus/CrealityV1 (RELOADED UI)_: π ∗ ( 1.75 / 2 ) 2 ∗ 5375 = 12928 mm 3 / s
> 
> _Velleman/K8400/Dual-head_ and _Velleman/K8400/Single-head_: π ∗ ( 1.75 / 2 ) 2 ∗ 1680 = 4041 mm 3 / s
> 
> _Wanhao/Duplicator i3 2.1_: π ∗ ( 1.75 / 2 ) 2 ∗ 5000 = 12026 mm 3 / s
> 
> So, to me, it seems reasonable to reduce these values somewhat. That said, I am not insisting the proposed changes must be accepted. The same goes for the recently added `SANITY_CHECK_E_STEPPER_RATES` macro in _SanityCheck.h_ (i.e., `#if ALL(HAS_EXTRUDERS, SANITY_CHECK_E_STEPPER_RATES)`). To me, `SANITY_CHECK_E_STEPPER_RATES` feels a bit overcautious. Anyway, should `//#define SANITY_CHECK_E_STEPPER_RATES` be added to _Configuration.h_ in this case.

I regularly print at 400mm/s (confirmed speed) with 0.5mm line width & 0.25mm line height. I'm fully aware that this is 50mm^3/sec, and that means my custom heatblock is working as it was designed to.
With that said, this is a **__volcano format__** hotend, not super-volcano, not goliath, and none of the really crazy multi-filament melters I've helped Monika with, or seen her working with others on. So I would in no way consider 400mm^3/sec as "not sane" ... though 12000mm^3/sec would basically require that you be printing houses with a concrete 3d printer, or similar, lol.

thinkyhead:
> I believe that even the adjusted values are still unrealistic.

The original Marlin config had E maximum rate at 10000 — a conspicuous five digit number — to indicate that the E steps were considered irrelevant in determining feedrate limits, particularly for the other (i.e., XY) axes. This also pertains to the speed of the E axis going in reverse where material volume doesn't matter. Contributors have maintained that custom.

> … should `//#define SANITY_CHECK_E_STEPPER_RATES` be added to `Configuration.h` …?

I only added that named option so I wouldn't be completely deleting the sanity checks, in case someone has a reason to use them. We don't necessarily want to expose everything that *could* be exposed as yet another user option because things like this will multiply forever and we are really at the breaking point with our historic approach to configuration.

AndreySamokhin:
Appreciate the clarification! That makes sense. Let's go ahead and drop this suggestion.

PR #1140, opened 2025-04-06
Fix babysteps and baudrate

photodude:
### Description

This patch corrects two issues which cause significant issues in setup and tuning of a BIBO printer.

1. enables babysteps which are needed for bltouch z-position corrections
2. baudrate was corrected to 115200

### Benefits

fixes inability to adjust z position when using bltouch as babysteps were not enabled
Fixes issues where M114 would not report the printer position due to baudrate being above 115200

Note: Users with factory mks tf28 screens will need to modify the screen baudrate to 115200. The factory baudrate of 250000 causes problems with gcode items which report back information like M114. Using 115200 on both the mainboard and the screen resolves this issue.

### DIFF
--- a/config/examples/BIBO/TouchX/default - BLTouch/Configuration.h
+++ b/config/examples/BIBO/TouchX/default - BLTouch/Configuration.h
@@ -96,7 +96,7 @@
  *
  * :[2400, 9600, 19200, 38400, 57600, 115200, 250000, 500000, 1000000]
  */
-#define BAUDRATE 250000
+#define BAUDRATE 115200
 
 //#define BAUD_RATE_GCODE     // Enable G-code M575 to set the baud rate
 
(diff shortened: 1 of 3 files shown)


thisiskeithb:
> Note: Users with factory mks tf28 screens will need to modify the screen baudrate to 115200. The factory baudrate of 250000 causes problems with gcode items which report back information like M114. Using 115200 on both the mainboard and the screen resolves this issue.

Please add this to a `README.md` file and include it with the configs since users won't know to check the PR queue for this information when upgrading their firmware breaks the serial connection.

photodude:
@thisiskeithb README.md added with note as requested. Should I also add a link to the MKS TF28 firmware?
I'll make one more commit to the USER instructions README.md to indicate the need to activate babysteps in the screen config as that's also part of the screen config changes

thisiskeithb:
> Should I also add a link to the MKS TF28 firmware?

Probably wouldn't hurt.

photodude:
@thisiskeithb Have done. This should be complete with the user notes now

PR #1143, opened 2025-04-20
🐛 Repack LCD tarballs without macOS resource files

### LINKED ISSUE #1142: [BUG] macOS resource files in tarballs break LCD updates
### Did you test the latest `bugfix-2.1.x` code?

Yes, and the problem still exists.

### Bug Description

https://github.com/MarlinFirmware/Configurations/archive/refs/tags/2.1.3-b2.zip

in Configurations-2.1.3-b2.zip\Configurations-2.1.3-b2\config\examples\Creality\Ender-3 V2\LCD Files - private.tar.gz

are 

._ files. preview/cache files created by macos or windows.

They need to be removed. I dont know if other LCD assets are also affected.

Dacai displays with private will error on the font.

Image

Image

### Bug Timeline

_No response_

### Expected behavior

LCD assets install.

### Actual behavior

LCD assets error out. See picture

### Steps to Reproduce

1. Load the mentioned private file.
2. Try to install on the display
3. See it error out on the font.

### Version of Marlin Firmware

probably more than one

### Printer model

Creality Ender 3 Max Neo / V2

### Electronics

_No response_

### LCD/Controller

Dacai

### Other add-ons

_No response_

### Bed Leveling

None

### Your Slicer

None

### Host Software

None

### Don't forget to include

- [x] A ZIP file containing your `Configuration.h` and `Configuration_adv.h`.

### Additional information & file uploads

Zip file with configs not needed.

thisiskeithb:
### Description

Remove macOS `._*` resource files from tarballs so LCD updates work correctly.

### Benefits

LCD updates work correctly.

### Related Issues

- https://github.com/MarlinFirmware/Configurations/issues/1142


thinkyhead:
Did you use a command like this…?
```sh
COPYFILE_DISABLE=1 tar --exclude='.DS_Store' --exclude='._*' -czf private.tar.gz private
```

thinkyhead:
Not super-important, but if you want to to preserve the old file permissions, 644 (rw-r--r--) instead of 755 (rwxr-xr-x) then you could fix that and then re-compress the tarballs as you did before.

PR #1144, opened 2025-06-09
Add E3D BigBox Dual Configuration Example

RainMotorsports:
### Requirements

Requires this PR adding support for this printer using PT100 amplifiers over the Rumba's EXP3 header from the factory. https://github.com/MarlinFirmware/Marlin/pull/27897

### Description

This configuration is for the E3D BigBox Dual in stock configuration. Tested on my personal machine with values checked against the original Marlin 1.1.0 RC configurations.

### DIFF
--- a/config/examples/E3D/BigBox Dual/README.md
+++ b/config/examples/E3D/BigBox Dual/README.md
@@ -1,0 +1,20 @@
+# Y AXIS HOMING WARNING
+If you get premature Y axis endstop triggers or even halts this is a known problem with the BigBox. The Y endstop wiring runs alongside the X motor wiring. Designer Greg Holloway recommends switching this setup to be active high and trigger low. My suggestion is to fix mount the Y axis switch to the frame and run the endstop wire down the front of the frame. 
+
+# Microstepping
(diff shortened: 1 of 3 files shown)


config/examples/E3D/BigBox Dual/README.md
```diff
@@ -0,0 +1,20 @@
+# Y AXIS HOMING WARNING
+If you get premature Y axis endstop triggers or even halts this is a known problem with the BigBox. The Y endstop wiring runs alongside the X motor wiring. Designer Greg Holloway recommends switching this setup to be active high and trigger low. My suggestion is to fix mount the Y axis switch to the frame and run the endstop wire down the front of the frame. 
+
+# Microstepping
+The Y axis should be jumpered to 8 microsteps with the stock geared setup with 0.9 degree stepper. 16 microsteps even at low speeds would overrun the RUMBA. 1.8 degree steppers can run 16 microsteps and there would be no change to the steps per mm in this configuration.
+
+#Probe
```
thisiskeithb: This is missing a space after `#` so it doesn’t render as a heading.

RainMotorsports: Fixed, since there was a mistake there I went ahead and fixed a non critical issue in the config I had noticed after running through a diff on my end. I had set the Z max to 50 when it was 100 in the original firmware. Should all be good to go pending the PR for the rumba pins.

PR #1154, opened 2025-09-25
Add config for Tronxy X5SA Pro in 2.1.3-b3

BeowulfNode42:
### Requirements

Tronxy X5SA Pro with TMC2225 drivers and Titan extruder and motherboard CXY-V6-191017 and chitu firmware era bootloader.

### Description

This is an initial setup for X5SA Pro specific configurations wtih the board CXY-V6-191017 and chitu firmware era bootloader. Not sure how this may need changing for later models that already come with a version of Marlin.

### Benefits

Fixes a few issues that are differences between the X5SA and Pro

- Homing all axes work due to pull request [#28059](https://github.com/MarlinFirmware/Marlin/pull/28059) and overrides in the configuration.h files here.
- Steppers no longer stop mid print.
- Acceleration speeds default to 500 rather than the 100 that the chitu firmware can handle.
- Linear Advance enabled.
- G29 actually works in Marlin, have selected Unified Bed Mesh Levelling. Whereas in the chitu firmware it measured and then performed the oposite action than what was needed.

### Related Issues

Initial push. Works on my printer with code from Marlin 2.1.3-beta3


thinkyhead:
Prepped and adjusted at #1158

PR #1156, opened 2025-09-25
fix: update TPARA config acording to new devs

chepo92:
### Description
Update TPARA: Three PArallelogram Robot Arm: A parallel manipulator with 3D printing toolhead, 

### Benefits
Fixes the basics for making the arm able to 3D print: G1 movements, homing procedures and calibration

Note: Movement is still jerky but may be improved in a future

### Related Issues
Acompanies  PR https://github.com/MarlinFirmware/Marlin/pull/28068


thinkyhead:
```diff
- // TPARA tower offset (position of Tower relative to bed zero position)
+ // TPARA Workspace offset relative to the tower (position of workspace origin relative to robot Tower origin)
```
Is this change in comment to indicate that the  `TPARA_OFFSET_*` is reversed (`workspace - tower` rather than `tower - workspace`) for all TPARA? If so, we will want to update the comment in all configurations.

How many of the TPARA configuration changes are always relevant? All relevant changes should be propagated to all example configurations, and the base configurations included with the Marlin source code.

thinkyhead:
If `HOME_Y_BEFORE_X` and/or `HOME_Z_FIRST` are required for `TPARA` then we can assert that in `SanityCheck.h` rather than duplicating these options inside the `AXEL_TPARA` block.

PR #1159, opened 2025-10-04
config: add Ender-3 Pro SKR Mini E3 BLTouch

gagath:
### Description

This configuration was created by copying the Ender-3 Pro SKR Mini E3/ configuration, and then enabling the BLTouch-related options from having a look at the diff between the CrealityV427/ and CrealityV427 - BLTouch/ configurations.

### Benefits

Allow a baseline for modded Ender-3 Pro with SKR Mini E3 **with BLTouch support**. Only the version without BLTouch is available currently.


ellensp:
BLTOUCH_FORCE_SW_MODE is not recommended, on cr touch and 3d touch this option is unreliable

thinkyhead:
@ellensp — _Very_ not recommended? 210edee06b

gagath:
Hi, thanks for updating the config to the latest standard. I did not know about `BLTOUCH_FORCE_SW_MODE` being not recommended as I just copied it. This PR now seems to disable it for other configs as well. I think it should be another PR as it looks out of scope? Let me know if there is anything I can do regarding this one.

I will rebuild the firmware and run the printer without `BLTOUCH_FORCE_SW_MODE` but I think it should work just fine.

gagath:
Tested OK on my Ender-3 Pro w/ BigTreeTech SKR Mini E3 3.0.

gagath:
Rebased on top of latest `import-2.1.x`. Ready to merge.

thinkyhead:
If turning off `BLTOUCH_FORCE_SW_MODE` for all these configs turns out to be a bad idea 210edee06b then we can fix it in a followup.

PR #1163, opened 2025-10-23
Exemple config for Ender 5 Plus with 4.2.2 GD32 mainboard

Gloom536:
### Requirements

Ender 5 Plus with 4.2.2 main board with GD32 chipset. TMC2208_Standalone Stepper Driver (H on the SD card port). Completly stock 3d Printer with no modification. Still has the OEM Touch Screen.

### Description
The actual exemple config are for the old 2.2 mainboard of the Ender 5 Plus, I bought my Ender 5 Plus and it came with a 4.2.2 Creality mainboard with a GD32F303RET6 chipset. I needed to change a lot of argument in the configuration.h and configuration_adv.h file to make it work. I made this in conjonction with nuck67d564a, thomastokaservermanwebhosting and dust9521on the Marlin discord server. They helped me a lot. I was a total noob with compiling stuff just before

### Benefits
Makes a Ender 5 Plus 4.2.2 GD32 mainboard work without hassle

### Related Issues
 Didn't see anyone requesting this except me


thinkyhead:
Could this machine be using the motherboard `BOARD_CREALITY_V422_GD32_MFL` instead?

According to the contributed pins file the standard 4.2.2 board only has 3 endstop connectors, and these are automatically mapped to MIN or MAX depending on your homing direction.

Does your GD32-based 4.2.2 board have 6 endstop plugs labeled as both MIN and MAX ?

ellensp:
"Could this machine be using the motherboard BOARD_CREALITY_V422_GD32_MFL instead"

 No.
1) Uarts need updated, GD32 starts from uart 0, stm32 from uart 1, so uarts need decremented by 1  
2) the MFL platform fails to boot with two uarts enabled  (ie with uart display DGUS_LCD_UI RELOADED)

Gloom536:
> Could this machine be using the motherboard `BOARD_CREALITY_V422_GD32_MFL` instead?
> 
> According to the contributed pins file the standard 4.2.2 board only has 3 endstop connectors, and these are automatically mapped to MIN or MAX depending on your homing direction.
> 
> Does your GD32-based 4.2.2 board have 6 endstop plugs labeled as both MIN and MAX ?

Yes, My ender 5 plus have v422 GD32 MFL board, but when I put this argument in the board section of configuration.h, there is no way to make the board boot. I tried with every "king pin" of the Marlin Discord. Only with the board I specified in this pull request I got it to work.

thinkyhead:
Ok, so the standard X, Y endstop plugs on a 4.2.2 board are PA5, PA6. Your machine is attaching its endstops to some plugs labeled "MIN" that connect to pins PA3, and PA14. We have no knowledge about a board with these specifics, and if this board differs in this way, it may differ in others, so we should get it formally defined in our pins pile.

ellensp:
"#define X_MAX_PIN  3 // Creality connects X_MAX to X_MIN(_PIN) motherboard connector
#define Y_MAX_PIN 14 // Creality connects Y_MAX to Y_MIN(_PIN) motherboard connector" 
There are left over from old config this was based on.

https://github.com/MarlinFirmware/Configurations/blob/import-2.1.x/config/examples/Creality/Ender-5%20Plus/CrealityV1%20(RELOADED%20UI)/Configuration.h#L76-L77

These pins are not even used on this board (and use wrong naming convention)

thinkyhead:
> these pins are not even used on this board (and use wrong naming convention)

Funky.

@Gloom536 are these the precise configs you used to build the firmware you are now running on your machine? The compiler should be barking that your endstops are being defined twice, but only the first definition actually counts, so the ones in the configuration (3, 14) should be getting applied, and if that is the case then we need to figure out just what the heck this board really is.

thinkyhead:
Ah, the pattern is clear. I've patched the other configurations that re-assign these pins too.

So, we just need to formally name this board and add a pins file for it that overrides these two endstop pins. If there are other differences compared to the standard 4.2.2 we can also apply them there.

```c
#define X_STOP_PIN PA3  // Creality connects X_MAX to X_MIN(_PIN) motherboard connector
#define Y_STOP_PIN PA14 // Creality connects Y_MAX to Y_MIN(_PIN) motherboard connector
```

thinkyhead:
Hmm… but in the other cases the board is `BOARD_RAMPS_CREALITY`. So, the pattern is even more clear. Pins 3, 14 only apply to RAMPS boards. And the endstop overrides in this submission are just an error.

Gloom536:
> > these pins are not even used on this board (and use wrong naming convention)
> 
> Funky.
> 
> @Gloom536 are these the precise configs you used to build the firmware you are now running on your machine? The compiler should be barking that your endstops are being defined twice, but only the first definition actually counts, so the ones in the configuration (3, 14) should be getting applied, and if that is the case then we need to figure out just what the heck this board really is.

The config I've put in this pull request is the one I use right now

thinkyhead:
>The config I've put in this pull request is the one I use right now

I hope you will enjoy using these updated merged versions!

Gloom536:
I've start to love my printer since I managed to update it with latest marlin build. I loved learning to compile a firmware too. And I'm happy if I can help people with the same mainboard than me to update their printer.

I've not followed all what you said earlier (I don't understand what all means lol). But is my firmware ok? Or should I update some setting to be 100% compatible? As of now I see no issue, but I'm not an expert at all.

thinkyhead:
> is my firmware ok?

Yes, I think you're good! Be sure to update to the release version when it comes out later this month, and give the new motion system a try. It makes the movement much smoother and quieter so you can push the speeds higher.

PR #1169, opened 2026-01-01
Addition of Monoprice Maker Ultimate 2

vernzimm:
This is for posterity mostly. I have since ported these into Marlin bugfix-2.1.x.

### Requirements

in /Marlin/
use/add included boards.h, pins.h files
add the pins_WEEDO_62B.h file

### Description

See readmes in the subfolders for origin/credit.


thinkyhead:
We can only accept PRs targeted to the `import-2.1.x` branch. Are you able to bring this up to date or do you need assistance with that?

thinkyhead:
Ah, I see your intent. Well, for posterity you can just maintain these things in your own fork and then direct all interested persons who would like to use Marlin 1.1.9 to that location. If Monoprice maintains a Marlin fork then they might want to bring some Marlin 1.1.8 that they distributed to their customers up to 1.1.9.

Anyway, as it is, this PR in its current form is not something that we could merge due to the inclusion of non-config files that belong to `MarlinFirmware/Marlin`, and PRs must be made merge-worthy by the contributor before we can accept them.

All that said, we will keep this around and the next time there's a compelling reason to update the `lts-1.1.9` branch I will try to remember to bring this stuff over at that time.

vernzimm:
@thinkyhead thanks for reviewing the situation anyway. I appreciate the feedback. It just feels like a shame to lose knowledge that is otherwise buried in some random company basement, but it really is outdated and was never good to begin with haha.

I am working on configs for bugfix-2.1.x. I also have board and pin files for that, but I'm still playing with them at the moment.

I'm currently struggling with UBL, and specifically the way that it stores the Z offset from probe, but at the same time it sort of maintains the Z offset from homing (the machine homes to Z Max, bed down, but has capacitive probe). I have found that toggling UBL on and off can cause Z to shift in one case, but not shift back in the other, moving the Z origin negative by ~0.6mm each time (I was not using baby stepping). It very quickly creates a situation where it would put the nozzle through the bed, so I have been doing everything with a 25.4mm length standard. I intuitively want the machine to operate relative to the probed and mesh offset origin, but that also creates the possibility that the bed software limit travel would be shifted beyond the Z Max home switch.

Any suggestions would be appreciated.

I also did finally realize that at some point I added Marlin in Discord, so I'm thinking that will be the best place to work on this. 🍻

PR #1176, opened 2026-02-13
Add stock Tevo Tarantula

ellensp:
### Requirements

a stock Tevo Tarantula

### Description

Based on MARLIN 2.0.8  from https://web.archive.org/web/20210506200100/https://github.com/JimBrown/MarlinTarantula/blob/EasyConfig-2.0.x/Marlin/Configuration.h

### Benefits

Now supported

### Related Issues

<li>MarlinFirmware/Marlin/issues/28335


ellensp:
Untested beyond building, based entirely on old configs

ellensp:
updated to match https://github.com/JimBrown/MarlinTarantula/blob/EasyConfig-2.0.x/Marlin/Configuration.h and https://web.archive.org/web/20230724095258/https://github.com/JimBrown/MarlinTarantula/blob/EasyConfig-2.0.x/Marlin/Configuration_adv.h (after extracting from json on broken page)

ellensp:
Users keep wanting the "easy config" version. this is not really compatible with marlin configuration examples
(It has some 60 custom options), but here it is updated for bugfix 
[TEVO TARANTULA EASY CONFIG Bugfix 2.1.x.zip](https://github.com/user-attachments/files/27969386/TEVO.TARANTULA.EASY.CONFIG.Bugfix.2.1.x.zip)

thinkyhead:
Possibly the `MAX_OVERSHOOT_PID_AUTOTUNE` option comes from the EZ config. Presumably by adding an array in front of `MOTHERBOARD` it will show a popup menu in ABM with only those options listed.

**Edit:** Ah, I see `MAX_OVERSHOOT_PID_AUTOTUNE` is a mystical option. Good, we need to keep the mystery alive.

PR #1183, opened 2026-03-22
Ender 5 Plus Configs - Increase Mesh Inset

### LINKED ISSUE #1106: [BUG] UBL doesn't probe right (X_MAX_POS) side of BED on Ender 5 Plus
### Did you test the latest `bugfix-2.1.x` code?

Yes, and the problem still exists.

### Bug Description

Stock Ender 5 Plus Printer with no changes / upgrades whatsoever. Downloaded the latest bugfix files and bugfix reloaded confirm for Ender 5 Plus. Made no changes to the code whatsoever.
Out of the box, when UBL is running, the firmware doesn't probe the right side (where X_MAX_POS endstop) is located while probes the rest of the bed just fine.
I googled the issue and I believe what others have suggested - the UBL is not able to reach the point.
I tried with the MESH_INSET option and the issue disappeared when I set it to 45 however, on the left side (X_MIN_POS) it was probing too far off the edge.
Finally, setting MESH_MAX_X to 45 gave the desired result.
Just putting this out here, incase other users are also facing this issue then, the Config files for E5P could be updated with these numbers.

### Bug Timeline

30/09/2024

### Expected behavior

Probes 25/25 points.

### Actual behavior

Probes 20/25 points leaving out the 5 points on the X_MAX side

### Steps to Reproduce

Stock printer
Default bugfix code
Default bugfix config

### Version of Marlin Firmware

Bugfix 2.1.X

### Printer model

Ender 5 Plus

### Electronics

Stock

### LCD/Controller

Stock

### Other add-ons

None

### Bed Leveling

UBL Bilinear mesh

### Your Slicer

Cura

### Host Software

Same as my slicer

### Don't forget to include

- [X] A ZIP file containing your `Configuration.h` and `Configuration_adv.h`.

### Additional information & file uploads

[Configuration_E5P.zip](https://github.com/user-attachments/files/17190194/Configuration_E5P.zip)

BLR-IIC:
As far as I can tell, the Ender 5 Plus configuration.h files have too low of a MESH_INSET.  This causes a couple problems:

### Description

1. UBL probing does not always probe unique bed positions and will probe the same spot more than once.
2. The power cord for the Y-carriage stepper motor is too short for the current MESH_INSET values and will be pulled out of its socket on the motor.

-->

### Benefits

Benefits of this PR include:

1. UBL probes unique locations on the print bed
2. The power cord for the Y-carriage stepper motor doesn't get stretched to the point its plug pulls out of the stepper motor's socket.

### Related Issues

May fix #1106

### DIFF
--- a/config/examples/Creality/Ender-5 Plus/BigTreeTech SKR 3/Configuration.h
+++ b/config/examples/Creality/Ender-5 Plus/BigTreeTech SKR 3/Configuration.h
@@ -2296,7 +2296,7 @@
 
   //#define MESH_EDIT_GFX_OVERLAY   // Display a graphics overlay while editing the mesh
 
-  #define MESH_INSET 45              // Set Mesh bounds as an inset region of the bed
+  #define MESH_INSET  1             // Set Mesh bounds as an inset region of the bed
   #define GRID_MAX_POINTS_X 5       // Don't use more than 15 points per axis, implementation limited.
   #define GRID_MAX_POINTS_Y GRID_MAX_POINTS_X
 
--- a/config/examples/Creality/Ender-5 Plus/BigTreeTech SKR 3/Configuration.h
+++ b/config/examples/Creality/Ender-5 Plus/BigTreeTech SKR 3/Configuration.h
@@ -2296,7 +2296,7 @@
 
   //#define MESH_EDIT_GFX_OVERLAY   // Display a graphics overlay while editing the mesh
 
-  #define MESH_INSET 1              // Set Mesh bounds as an inset region of the bed
+  #define MESH_INSET 45              // Set Mesh bounds as an inset region of the bed
   #define GRID_MAX_POINTS_X 5       // Don't use more than 15 points per axis, implementation limited.
   #define GRID_MAX_POINTS_Y GRID_MAX_POINTS_X
 
--- a/config/examples/Creality/Ender-5 Plus/BigTreeTech SKR 3/Configuration_adv.h
+++ b/config/examples/Creality/Ender-5 Plus/BigTreeTech SKR 3/Configuration_adv.h
@@ -2538,8 +2538,8 @@
   // Override the mesh area if the automatic (max) area is too large
   #define MESH_MIN_X MESH_INSET
   #define MESH_MIN_Y MESH_INSET
-  #define MESH_MAX_X X_BED_SIZE - (MESH_INSET)
-  #define MESH_MAX_Y Y_BED_SIZE - (MESH_INSET)
+  #define MESH_MAX_X (X_BED_SIZE - 44 - (MESH_INSET))
+  #define MESH_MAX_Y (Y_BED_SIZE -  5 - (MESH_INSET))
 #endif
 
 #if ALL(AUTO_BED_LEVELING_UBL, EEPROM_SETTINGS)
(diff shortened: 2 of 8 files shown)


Patronics:
I can confirm that at least for my Ender 5 Plus with MicroSwiss NG hotend upgrade, and running Marlin 2.1.3-beta 3, this adjustment fixed probing issues, where otherwise the probe was off the edge of the bed for some probes, leading to an early end to the probing process.

BLR-IIC:
> I can confirm that at least for my Ender 5 Plus with MicroSwiss NG hotend upgrade, and running Marlin 2.1.3-beta 3, this adjustment fixed probing issues, where otherwise the probe was off the edge of the bed for some probes, leading to an early end to the probing process.

I'm happy to see it works for you :)  For reference, my Ender 5 Plus has the stock motors, carriage, gantry layout and equipment.

thinkyhead:
This will give you more complete bed coverage:
```c
  #define MESH_MIN_X 1
  #define MESH_MIN_Y 1
  #define MESH_MAX_X (X_BED_SIZE - 44 - 1)
  #define MESH_MAX_Y (Y_BED_SIZE - 5 - 1)
```
The values 44 and 5 assume the nozzle cannot move outside the bed. There may be more actual freedom of movement on the right and towards the back, so you should test that to see if a slightly larger area of the bed can be incorporated.

BLR-IIC:
If I understand the code correctly, MESH_MAX_Y is still too small.  The Y-carriage motor power cord will pull right out of the motor somewhere 25 mm from the edge of the bed measured from the nozzle.  This is a physical limitation of the stock equipment.  I can't remember what the Y-offset of the probe is off-hand, and my printer isn't in a good spot where I could measure relative to the probe (it's the stock location of the BL Touch probe - I swapped a CR Touch in early last year but I believe the offsets should still be the same).

I suggest an offset that would keep the nozzle at least 30 mm from the Y-edge of the print bed, maybe even 35 mm for some extra safety (I'm going to stick with my 45 mm number personally - "If it ain't broke...").

thinkyhead:
Surely not. A 3D printer must be able to physically move the nozzle to all four corners of the bed as reckoned relative to the endstops, otherwise the bed size is just wrong. So what is the true bed size on this machine, if it cannot be allowed to move beyond some point?

BLR-IIC:
The physical bed size is 375 mm in the X and 372 mm in the Y.  Assuming X=0 and Y=0 is the lower-left corner of the bed, Y=0 has to be inset enough so that cords aren't pulled out from anywhere.  Therefore the actual printing area is a bit smaller than the physical bed size.  Also, the bed's heater loses effectiveness at the edges, but that's beside the topic here.

Clarification: It's the X-axis motor on the Y-carriage that is losing the power cord due to being pulled out with the inset values currently in the repo.  I discovered this was happening after getting some homing errors showing no movement in the X-axis, only to find the cord was being pulled out with the repo's inset values.

If it would be helpful, I can provide some measurements from my machine to help inform a functional change that won't physically break anything - let me know what's needed.

thinkyhead:
If your X axis cannot move to X0 then your X0 is too far to the left. Since this machine homes to MAX on X and Y this effectively means your X bed size is not 375mm. It must be reduced by whatever distance puts your X0 at a point where the cable does not pull out.

This may not be an issue with every build of this machine, so we need to verify with others whether this is a common flaw in the design.

PR #1184, opened 2026-03-27
Support BLTouch V.1.3 with pinboard adapter on Ender-3 Pro with 4.2.7 mainboard

joriskleiber:
### Description

I'm using this ancient BLTouch V.1.3 with my Ender 3 Pro and the v4.2.7 mainboard.

As you can see the wiring of this old kit doesn't work with the new 5 pin port on the v4.2.7 mainboard because voltage and ground are switched.
So I kept using the pinboard but there was no option tell marlin that I'm using this pinboard.
BLTouch_V 1 3
NewUpgradeSilentMainboardV4 2 7

Maybe these changes should also be ported to other Creality printers (Ender 3, Ender 3 Pro, Ender 5, CR-10 Mini, CR-10)  but I don't feel confident creating an BLTouch config for them.

### Benefits

Adding/enabling these options allows me to keep using the pinboard and BLTouch without having to rewire it.

### DIFF
--- a/config/default/Configuration.h
+++ b/config/default/Configuration.h
@@ -1015,7 +1015,9 @@
 
 // @section polargraph
 
-// Enable for Polargraph Kinematics
+/**
+ * Polargraph Kinematics
+ */
 //#define POLARGRAPH
 #if ENABLED(POLARGRAPH)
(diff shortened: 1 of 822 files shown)


thinkyhead:
Merged as commit 46f7846a73

PR #1192, opened 2026-05-30
fix(Configurations): rename RESONANCE_TEST to FTM_RESONANCE_TEST (411 occurrences)

WakinCode:
### Description

Global rename was missed in the Configurations repo when FTM_RESONANCE_TEST 
was introduced in the main Marlin repo. All 411 occurrences were bare 
RESONANCE_TEST with no existing FTM_RESONANCE_TEST references present.

Replaced via:
grep -rlZ 'RESONANCE_TEST' . | xargs -0 sed -i 's/\bRESONANCE_TEST\b/FTM_RESONANCE_TEST/g'

Tested: zero remaining bare RESONANCE_TEST occurrences confirmed by grep.
Affects: config/default and all config/examples across all printer variants.

### Benefits

Enables FTM_RESONANCE_TEST to compile correctly when uncommented in any 
example config. Previously the bare RESONANCE_TEST define was silently 
ignored by the preprocessor, causing the feature to appear configured 
but never actually compile in.

### Related Issues

None — discovered while testing FTM_RESONANCE_TEST on Ender 3 V2 with 
Creality 4.2.2 board.


thinkyhead:
This branch contains provisional changes that are not yet merged in Marlin / bugfix-2.1.x. Will merge the PR over there shortly.

PR #1199, opened 2026-08-09
add Ender-3 Pro + BTT_SKR_MINI_E3_V2_0 + bltouch

ellensp:
### Requirements

Ender-3 Pro + BTT_SKR_MINI_E3_V2_0 + bltouch

### Description

Helped out a user in Discord to create this config; I didn't want it to just vanish, so I created this PR

Based on config/examples/Creality/Ender-3 Pro/BigTreeTech SKR Mini E3 3.0/BLTouch and modified for a  BTT_SKR_MINI_E3_V2_0

Switched to AUTO_BED_LEVELING_BILINEAR so the firmware would fit in an STM32F103RC

Has to create subdirectories "base" for the original config and added directory "bltouch" for the new config 

### Benefits

This example configuration exists


thinkyhead:
Comparing to `config/examples/Creality/Ender-3 Pro/BigTreeTech SKR Mini E3 2.0/base` there are some significant configuration differences apart from the probe and leveling-related changes. Some of these might be needed to save space. Some might be useful if they can be enabled and made to fit by shaving other options.

Compare this set to the base set and cherry-pick the changes from this into to the `base` config, then cherry-pick changes from the `base` config into the `BLTouch` config. Copy-paste of configs onto configs plus GitHub Desktop should be all that are needed to perform this light surgery.

PR #1201, opened 2026-08-12
[Sovol][SV-06] Homing Bump of XY should be 0 because of sensorless probing

EvansJahja:
### Requirements

### Description

Because of sensorless probing, HOMING_BUMP_MM of X and Y axis should be 0 per documentation. I have found that my printer experience less grinding at the end-stop because of this as well.

### Benefits

Improve homing, reduces grinding

### Related Issues

-


thinkyhead:
Combined with #1197

